PR #7, opened 2022-03-07
v0.0.2 Integration Testing

miwright2:
- Integration testing in preparation for the v0.0.2 build.
- Resolve issue when resuming upload after a disconnection (CMB-APPS-52)
- Documentation updates.

### DIFF
--- a/combustion-android-ble/src/main/java/inc/combustion/framework/ble/ProbeManager.kt
+++ b/combustion-android-ble/src/main/java/inc/combustion/framework/ble/ProbeManager.kt
@@ -148,7 +148,14 @@ internal open class ProbeManager (
         addJob(owner.lifecycleScope.launch(Dispatchers.IO) {
             while(isActive) {
                 if(isConnected.get() && mac != SimulatedProbeManager.SIMULATED_MAC) {
-                    remoteRssi.set(peripheral.rssi())
+                    try {
+                        remoteRssi.set(peripheral.rssi())
+                    } catch (e: Exception) {
+                        if (DebugSettings.DEBUG_LOG_BLE_OPERATIONS) {
+                            Log.d(LOG_TAG,
+                                "Ignoring exception reading remote RSSI: \n${e.stackTrace}")
+                        }
+                    }
                     _probeStateFlow.emit(probe)
                 }
                 delay(PROBE_REMOTE_RSSI_POLL_RATE_MS)
@@ -258,8 +265,15 @@ internal open class ProbeManager (
 
     private suspend fun readFirmwareVersion() {
         withContext(Dispatchers.IO) {
-            val fwVersionBytes = peripheral.read(FW_VERSION_CHARACTERISTIC)
-            fwVersion = fwVersionBytes.toString(Charsets.UTF_8)
+            try {
+                val fwVersionBytes = peripheral.read(FW_VERSION_CHARACTERISTIC)
+                fwVersion = fwVersionBytes.toString(Charsets.UTF_8)
+            } catch (e: Exception) {
+                if (DebugSettings.DEBUG_LOG_BLE_OPERATIONS) {
+                    Log.d(LOG_TAG,
+                        "Ignoring exception reading remote FW version: \n${e.stackTrace}")
+                }
+            }
         }
     }
 
--- a/combustion-android-ble/src/main/java/inc/combustion/framework/ble/ProbeManager.kt
+++ b/combustion-android-ble/src/main/java/inc/combustion/framework/ble/ProbeManager.kt
@@ -151,10 +151,8 @@ internal open class ProbeManager (
                     try {
                         remoteRssi.set(peripheral.rssi())
                     } catch (e: Exception) {
-                        if (DebugSettings.DEBUG_LOG_BLE_OPERATIONS) {
-                            Log.d(LOG_TAG,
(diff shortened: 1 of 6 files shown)


combustion-android-ble/src/main/java/inc/combustion/framework/ble/ProbeManager.kt
```diff
@@ -148,7 +148,14 @@ internal open class ProbeManager (
         addJob(owner.lifecycleScope.launch(Dispatchers.IO) {
             while(isActive) {
                 if(isConnected.get() && mac != SimulatedProbeManager.SIMULATED_MAC) {
-                    remoteRssi.set(peripheral.rssi())
+                    try {
+                        remoteRssi.set(peripheral.rssi())
+                    } catch (e: Exception) {
+                        if (DebugSettings.DEBUG_LOG_BLE_OPERATIONS) {
+                            Log.d(LOG_TAG,
```
nick-sasquatch, line 155: What kind of exceptions were you seeing here? Is `debug` the right level of logging? Is there one that we can explicitly ignore? If so, it might make sense to catch that type of exception rather than the generic `Exception`.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/ProbeManager.kt
```diff
@@ -258,8 +265,15 @@ internal open class ProbeManager (
 
     private suspend fun readFirmwareVersion() {
         withContext(Dispatchers.IO) {
-            val fwVersionBytes = peripheral.read(FW_VERSION_CHARACTERISTIC)
-            fwVersion = fwVersionBytes.toString(Charsets.UTF_8)
+            try {
+                val fwVersionBytes = peripheral.read(FW_VERSION_CHARACTERISTIC)
+                fwVersion = fwVersionBytes.toString(Charsets.UTF_8)
+            } catch (e: Exception) {
```
nick-sasquatch, line 271: See note about exception handling above.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/ProbeManager.kt
```diff
@@ -148,7 +148,14 @@ internal open class ProbeManager (
...
                 if(isConnected.get() && mac != SimulatedProbeManager.SIMULATED_MAC) {
-                    remoteRssi.set(peripheral.rssi())
+                    try {
+                        remoteRssi.set(peripheral.rssi())
+                    } catch (e: Exception) {
+                        if (DebugSettings.DEBUG_LOG_BLE_OPERATIONS) {
+                            Log.d(LOG_TAG,
+                                "Ignoring exception reading remote RSSI: \n${e.stackTrace}")
+                        }
+                    }
```
nick-sasquatch, line 158: What kind of exceptions were you seeing here? Is `debug` the right level of logging? Is there one that we can explicitly ignore? If so, it might make sense to catch that type of exception rather than the generic `Exception`.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/ProbeManager.kt
miwright2: See comment below.

combustion-android-ble/src/main/java/inc/combustion/framework/log/Session.kt
```diff
@@ -68,6 +68,22 @@ internal class Session(seqNum: UInt, private val serialNumber: String) {
     val isEmpty get() = _logs.isEmpty()
     val maxSequenceNumber: UInt get() = if(isEmpty) 0u else _logs.lastKey()
 
+    val maxSequencialSequenceNumber: UInt get() {
```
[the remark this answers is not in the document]
miwright2: Fixed

combustion-android-ble/src/main/java/inc/combustion/framework/log/Session.kt
```diff
@@ -136,9 +152,15 @@ internal class Session(seqNum: UInt, private val serialNumber: String) {
                 // don't change the next expected
             }
             else {
-                Log.w(LOG_TAG,
-                    "Received unexpected record? " +
-                            "$serialNumber.${logResponse.sequenceNumber} ($nextExpectedRecord)")
+                // the following can occur when re-requesting records to handle gaps in the
+                // record log.
+                if(DebugSettings.DEBUG_LOG_TRANSFER) {
```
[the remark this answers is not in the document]
miwright2: I agree -- with the suggestion.  Will incorporate outside of this PR when timing makes sense.  Will keep the flags the same for use by calling apps.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/ProbeManager.kt
miwright2: I noticed BLE disconnect exceptions while the RSSI read operation was in progress when doing range testing.  The exception is from the `kable` library, but didn't quite match their docs.  

I'm going to remove the conditional logging and log the exception as a warning.  I think a BLE exception is a normal case in operation, but want to handle all exceptions (and hence the warning), since we should really not crash in this case, and the best we can do is log a warning message and continue without updating data.

PR #9, opened 2022-04-04
Update minSdk to 24

angrygorilla:
updated minSdk to 24 which uncovered a couple API incompatibilities. 

- There was a bluetooth isScannable method that was introduced in API 26, to handle this we just return true for API versions less than 26
- There are several time/date/formatting methods that were added in Java 8 that we were using. To work around that I am using some older java.util and java.text modules to create some timestamps for session id creation.  

There are some issues testing this min version with our app in a emulator currently.

### DIFF
--- a/combustion-android-ble/build.gradle
+++ b/combustion-android-ble/build.gradle
@@ -8,7 +8,7 @@ android {
     compileSdk 31
 
     defaultConfig {
-        minSdk 28
+        minSdk 24
         targetSdk 31
         versionCode 1
         versionName "0.0.3"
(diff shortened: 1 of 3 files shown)


miwright2 (approved):
- Change looks good.  Noted Kotlin optional Kotlin style comment in Slack.
- Tested record transfer using Diagnostic App.  Works great with `0.5.0` device firmware.
- If there is an issue with `0.6.1` firmware requiring Android change to `DeviceStatus` deserialization, then let's consider adding to this PR.

Approved based on outcome of above.

PR #11, opened 2022-05-13
Read HW revision from probe

jjohnstz:


miwright2 (approved):
- Tested the branch against `develop` of the diagnostic app in `combustion-android` using v0.4 dev board and a v0.6.1 build off of develop from probe firmware.
- PR looks good.
- Suggest updating the Gradle Plugin and Kotlin versions as part of this PR.

PR #12, opened 2022-05-18
Decode multiple UART responses per notification

jjohnstz:


miwright2 (approved):
This looks good.

Test Configuration
- I used the CIP-20 feature branch on the firmware w/ my v0.4 dev board.
- I also have some old v0.4 firmware on another dev board.
- I used develop of the Engineering App
- And this feature branch.

I was able to connect to and upload from both probes with no issues.  The new firmware does look faster.

PR #13, opened 2022-05-28
Instant Read, Probe ID, Probe Color, Timer Support

miwright2:
- Support Instant Read Mode Device Status and Advertising Packets.
- Read Probe ID and Probe Color.  UART commands to set ID and color are pending.
- Filter out timers when publishing Combustion Devices to the Discovered Probes Flow.
- Fixes for CMB-APPS-60 and CMB-APPS-58.

Please use `feature/instant_read_modecolor_timer` branch in `combustion-android` when testing.


jjohnstz (approved):
Looks good

PR #14, opened 2022-06-01
Foreground Service & ProbeScanner

miwright2:
- Ability to optionally launch and run the Combustion Service as a foreground service.
- Calling app is responsible for creating the `Notification` and passing that into the service on start.
- Passing in null for the `Notification` results in the service being run in the background.
- Add `ProbeScanner` and `ProbeScanResult` to the public framework API to allow calling apps to scan for and identify Combustion probes independent of the Combustion Service.

Please test with `feature/foreground-service_and_scanning` in `combustion-android` repo.  Related PR is [here](https://github.com/combustion-inc/combustion-android/pull/26)  Example app updates are pending.

### DIFF
--- a/.idea/gradle.xml
+++ b/.idea/gradle.xml
@@ -14,7 +14,6 @@
             <option value="$PROJECT_DIR$/combustion-android-ble" />
           </set>
         </option>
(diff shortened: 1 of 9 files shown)


jjohnstz (approved):
Looks good

PR #17, opened 2022-06-10
Suspend Version of ProbeScanner

miwright2:
- Making suspending portion of scanning public in ProbeScanner.
- Users of the API are required to manage the coroutine scope for the suspend call.


jjohnstz (approved):
Looks good

PR #18, opened 2022-06-14
feat(battery status): Add battery status processing

nick-sasquatch:


combustion-android-ble/src/main/java/inc/combustion/framework/ble/DeviceStatus.kt
```diff
@@ -70,13 +73,22 @@ internal data class DeviceStatus(
             val probeID = if(modeColorId != null) ProbeID.fromUByte(modeColorId) else ProbeID.ID1
             val probeMode = if(modeColorId != null) ProbeMode.fromUByte(modeColorId) else ProbeMode.Normal
 
+            // use status if available
+            val status = if (data.size > 22)
+                data.sliceArray(STATUS_RANGE)[0]
+            else
+                null
+
+            val batteryStatus = if(status != null) ProbeBatteryStatus.fromUByte(status) else ProbeBatteryStatus.Ok
```
miwright2, line 82: Just a note, you can use the [Elvis operator](https://kotlin-quick-reference.com/156-R-elvis-operator.html) to turn this line into a statement instead of a conditional.  

I probably should have made the same comment above on `probeID` and `probeMode`.  I'm easy on the approach, just want to point out.

miwright2 (approved):
I noticed that prior to this PR, there is inconsistency in the style for `enum class` values (all caps versus camel case).  If you are willing [to review and cherry-pick this commit](https://github.com/combustion-inc/combustion-android-ble/commit/f0cd690abbbbceee95fd7831f1b89395142f9fc4) into this branch that would be great.  Otherwise, I can set up another PR.

Otherwise, everything looks good.  
- I tested with PR #29 in `combustion-android` and PR #96 in `meat-dagger-firmware`.  Able to see `OK` battery status.  
- Didn't test with a power supply for `LOW_BATTERY` condition, assuming that has already been covered.  
- Also tested with `0.8.0_rc3` firmware on form factor probe, and default value (`OK`) is displayed.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/DeviceStatus.kt
nick-sasquatch: It'd be nice to be consistent--I'm happy to make the changes to `probeID` and `probeMode`. What do you think?

PR #19, opened 2022-07-11
Feature: session info

jjohnstz:


combustion-android-ble/src/main/java/inc/combustion/framework/log/ProbeTemperatureLog.kt
```diff
@@ -42,76 +42,50 @@ import java.util.*
  */
 internal class ProbeTemperatureLog(private val serialNumber: String) {
 
-    private val sessions: SortedMap<SessionId, Session> = sortedMapOf()
-    var currentSessionId = SessionId.NULL_SESSION_ID
-        private set
+    private var sessions: Array<Session> = arrayOf()
```
miwright2, line 45: Makes sense

miwright2 (approved):
Looks good.

I tested the Diagnostic and Example apps with this version.  I was able to download two different probes with both.

PR #21, opened 2022-07-27
Export Data as CSV

miwright2:
- Ability to export data as CSV.
- Add in desugar library for newer DateTime calls.
- Fix issue with flow replay buffer.
- Setup versioning based on git describe.

Please test against this branch of the open source app:
https://github.com/combustion-inc/combustion-android-example/pull/13

### DIFF
--- a/combustion-android-ble/build.gradle
+++ b/combustion-android-ble/build.gradle
@@ -4,6 +4,20 @@ plugins {
     id("maven-publish")
 }
 
+def getVersionName = { ->
+    try {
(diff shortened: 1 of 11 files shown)


combustion-android-ble/build.gradle
```diff
@@ -4,6 +4,20 @@ plugins {
     id("maven-publish")
 }
 
+def getVersionName = { ->
```
angrygorilla, line 7: 👊🏻

angrygorilla (approved):
great work adding this change so quickly!

PR #22, opened 2022-08-17
Add CRC Calculation and Checking to UART Messages

miwright2:
- Check CRC on response messages.
- Calculate and send CRC on request messages.
- Firmware greater than v0.8.2 requires this feature.

Tested data transfer, set probe id, and probe color.


nick-sasquatch (approved):
Looks good!

PR #24, opened 2022-09-15
Preliminary Prediction Support

miwright2:
- Support deserializing message updates for prediction.
- Updates to `Probe` data class and additional enums in public packages.


combustion-android-ble/src/main/java/inc/combustion/framework/ble/uart/SetColorResponse.kt
```diff
@@ -1,5 +1,35 @@
+/*
```
jjohnstz, line 1: Thanks for adding this in.

jjohnstz (approved):

PR #27, opened 2022-10-04
Prediction Message Forwards Compatibility

miwright2:
- Cherry-pick code from develop to use the packet length in the response when processing UART response messages.

Tested with probe firmware `v0.8.2` and `v0.9.0`


jjohnstz (approved):
Looks good

PR #28, opened 2022-10-11
Set Prediction

miwright2:
- API for setting and canceling a prediction.
- Change Warming State to Cooking State.

Requires this Probe FW
https://github.com/combustion-inc/meat-dagger-firmware/pull/128

Requires this Android Example App
https://github.com/combustion-inc/combustion-android-example/pull/20


jjohnstz (approved):
Looks good

PR #29, opened 2022-12-02
DROID-23: Android 13 Support

angrygorilla:
- Updates for Android 13, updated dependencies
- Removed depreciated kotlin flag `useIR`
- Updated Kotlin version to match the example app


miwright2 (commented):
Can you update the README to reference that `POST_NOTIFICATION` is a required permission.

PR #31, opened 2023-01-06
DROID-62: Improve Probe Data Class

miwright2:
- Have probe data class behave more like iOS

### DIFF
--- a/combustion-android-ble/src/main/java/inc/combustion/framework/ble/ProbeManager.kt
+++ b/combustion-android-ble/src/main/java/inc/combustion/framework/ble/ProbeManager.kt
@@ -523,11 +523,11 @@ internal open class ProbeManager (
             predictionStatus = probeStatus?.predictionStatus
 
             // handle large predictions and prediction resolution
-            if(predictionStatus != null) {
-                val rawPrediction = predictionStatus!!.predictionValueSeconds
+            predictionStatus?.let {
+                val rawPrediction = it.predictionValueSeconds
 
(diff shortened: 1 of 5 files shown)


combustion-android-ble/src/main/java/inc/combustion/framework/ble/ProbeManager.kt
```diff
@@ -488,6 +489,112 @@ internal open class ProbeManager (
...
+        else if(mode == ProbeMode.NORMAL) {
+            val virtualSensors = probeStatus?.virtualSensors ?: advertisingData.virtualSensors
+            val instantReadCelsius = if(!instantReadMonitor.isIdle(PROBE_INSTANT_READ_IDLE_TIMEOUT_MS)) _probe.instantReadCelsius else null
+
+            predictionStatusMonitor.activity()
+
+            predictionStatus = probeStatus?.predictionStatus
+
+            // handle large predictions and prediction resolution
+            if(predictionStatus != null) {
```
jjohnstz, line 526: Nitpick: I think it would be safer to use let:
`predictionStatus?.let {
    // use the prediction status
}`

jjohnstz (approved):
Looks good. One minor comment.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/ProbeManager.kt
miwright2: good call.  got it.

PR #32, opened 2023-01-12
DROID-74: framework ble package refactor

miwright2:


combustion-android-ble/src/main/java/inc/combustion/framework/ble/BleDevice.kt
```diff
@@ -0,0 +1,253 @@
...
+ * @param adapter Android BluetoothAdapter.
+ */
+internal open class BleDevice (
+    private val mac: String,
+    protected val owner: LifecycleOwner,
+    var advertisingData: LegacyProbeAdvertisingData,
+    adapter: BluetoothAdapter
+) {
+    companion object {
+        private const val PROBE_REMOTE_RSSI_POLL_RATE_MS = 2000L
```
nick-sasquatch, line 62: If these are the same across all devices, I'd suggest removing the `PROBE_` prefix.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/BleDevice.kt
```diff
@@ -0,0 +1,253 @@
...
+                // engine = SystemLogEngine
+                // level = Logging.Level.Events
+                // format = Logging.Format.Multiline
+                // data = Hex
+                 */
+            }
+        }
+
+    protected val connectionMonitor = IdleMonitor()
+    val remoteRssi = AtomicInteger(0)
```
nick-sasquatch, line 96: Spoke with Matt about this--please put a `TODO` in here to make private when advertising data injection's addressed.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/BleDevice.kt
```diff
@@ -0,0 +1,253 @@
...
+                }
+            }
+        })
+    }
+
+    fun observeRemoteRssi(callback: (suspend (rssi: Int) -> Unit)? = null) {
+        jobManager.addJob(owner.lifecycleScope.launch(Dispatchers.IO) {
+            var exceptionCount = 0;
+            while(isActive) {
+                if(isConnected.get() && mac != SimulatedProbeManager.SIMULATED_MAC) {
```
nick-sasquatch, line 126: Can the simulated probe be handled closer to the probe device implementation?

nick-sasquatch (approved):
Just a couple minor comments--approving pending comments being addressed. Nice work.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/BleDevice.kt
miwright2: I hope this captures the design intent for when scanning is integrated.

```
    // TODO: This variable should ideally be scoped as private.  
    //  This should be possible when this class is able handle and process advertising
    //  packets directly from the device scanner.
```

combustion-android-ble/src/main/java/inc/combustion/framework/ble/BleDevice.kt
miwright2: I think so.  I put together the following design notes under DROID-62 that I think would cover that... I'll start a thread in Slack for opinions.

image

PR #36, opened 2023-01-20
DROID-77: Data stale flags

jjohnstz:


angrygorilla (approved):
Changes look good. We will revisit any funny issues when module refactor is complete. Capture in ticket https://linear.app/combustion/issue/DROID-85/stale-data-detection-stale-data-indication-is-taking-longer-than

PR #38, opened 2023-01-31
DROID-91: MeatNet connection management

miwright2:
- Create DataLinkArbitrator
- Heuristics to determine if a device should be connected to or disconnected from.
- Connection and Disconnection
- Read Device Information on Connection and publish.
- Flow for publishing node firmware details upon connection.
- MeatNet: automatic connection and reconnection to timers.
- Auto-Reconnect: !MeatNet with autoReconnect = true, automatic reconnect after framework call to connect.  Continue to reconnect until framework call disconnect.

### DIFF
--- a/combustion-android-ble/src/main/java/inc/combustion/framework/ble/DataLinkArbitrator.kt
+++ b/combustion-android-ble/src/main/java/inc/combustion/framework/ble/DataLinkArbitrator.kt
@@ -157,11 +157,12 @@ internal class DataLinkArbitrator(
             // if not using meatnet, and the client app is asking us to connect
             if(fromApiCall) {
                 device.shouldAutoReconnect = settings.autoReconnect
+                return canConnect
             }
 
             // if not using meatnet, and this is a call from inside the framework then
             // we want to use the auto-reconnect setting.
-            return canConnect
+            return device.shouldAutoReconnect
         }
 
         // if not meatnet and not a probe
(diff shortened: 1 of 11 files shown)


combustion-android-ble/src/main/java/inc/combustion/framework/service/DeviceManager.kt
```diff
@@ -497,6 +497,13 @@ class DeviceManager(
      */
     fun stopDfuMode() = service.stopDfuMode()
 
+    /**
+     * State flow containing the firmware details for all nodes on the network.
+     */
+    fun getNetworkFirmwareState(): StateFlow<FirmwareState>? {
```
miwright2, line 503: @nick-sasquatch -- this adds the flow for firmware information for network to be used on the device list screen.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/DataLinkArbitrator.kt
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    fun shouldUpdateOnOutOfRange(device: ProbeBleDeviceBase): Boolean {
+        // TODO: Implement Multiplexing Business Logic
+        return debuggingWithStaticLink(device)
+    }
+
+    fun shouldHandleAdvertisingPacket(device: ProbeBleDeviceBase): Boolean {
+        /*
+        return true
```
jjohnstz, line 204: Is this still needed?  Maybe add a TODO comment

jjohnstz (approved):
Looks great.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/device/BleDevice.kt
```diff
@@ -152,6 +163,7 @@ internal open class BleDevice (
                     }
 
                     isConnected.set(connectionState == DeviceConnectionState.CONNECTED)
+                    isDisconnected.set(state is State.Disconnected)
```
nick-sasquatch, line 166: I don't see any usages where `isConnected` and `isDisconnected` checks could get interleaved, but if there really is some `Atomic`-ness necessary here, I'd make sure that there's not something that could sneak in between these two and end up with both being true.

combustion-android-ble/src/main/java/inc/combustion/framework/service/DeviceManager.kt
nick-sasquatch: Thanks for getting this in there!

nick-sasquatch (approved):
Thanks for getting the firmware update checks in! I left the rest of the review for Jesse with the exception of a minor comment about something I saw scrolling down to the DFU stuff.

PR #42, opened 2023-02-03
Droid-101: Read firmware version for timer without probes

jjohnstz:


combustion-android-ble/src/main/java/inc/combustion/framework/ble/NetworkManager.kt
```diff
@@ -398,16 +397,49 @@ internal class NetworkManager(
...
+                else {
+                    if(firmwareStateOfNetwork[advertisingData.id] == null) {
+                        handleTimerWithoutProbes(advertisingData)
+                    }
+                }
+            }
+        }
+    }
+
+    private fun handleTimerWithoutProbes(advertisement: CombustionAdvertisingData) {
```
nick-sasquatch, line 418: This probably ought to be `handle[Repeater|MeatNetNode|etc]WithoutProbes` or whatever terminology we're using for that class of device. I can pick that up in a later PR if you'd like.

nick-sasquatch (approved):
One nitpick that we can pick up later. Nice work.

PR #46, opened 2023-02-07
DROID-13: Add DFU support to framework, including Nordic library integration

nick-sasquatch:
This also addresses DROID-58 and DROID-109.


jjohnstz (approved):
Looks great

PR #47, opened 2023-02-07
Add delay to start of prediction status monitor

jjohnstz:


miwright2 (approved):
Looks good. If possible, can we change the name of `predictionStale` to match iOS.

PR #51, opened 2023-02-17
DROID-132: Do not ignore status notification if prediction has changed

jjohnstz:


miwright2 (approved):
Looks good.  Nice catch on these.  Approved.

PR #52, opened 2023-02-20
DROID-33: Improvements to Service Lifecycle Management

miwright2:
- Make NetworkManager a singleton that is initialized when the service is started.
- Correctly clear shared flow when devices are cleared.


angrygorilla (approved):
Tested out changes on Pixel 7. Tested all of the enable/disable and theme switched and hit no issues. Nice changes!

PR #53, opened 2023-02-21
DROID-126, DROID-123: Address ANR

miwright2:
- Run collect on Dispatchers.IO, not main thread.
- Make the emit call non-blocking.


nick-sasquatch (approved):
I've tried this and haven't seen any issues. The approach looks good to me. Nice work.

PR #56, opened 2023-02-23
DROID-4: CSV v4 Support

miwright2:
- Processes Prediction Log structures and add them to the temperature log.
- Update CVS to Version 4 
- Stop supporting pre-production protocol versions.


nick-sasquatch (approved):
Checked mostly for data flow continuity between modules--only spot checks for sanity on raw data conversion (happy to double-check some of the stuff in the weeds if you'd like, though). Looks good!

PR #59, opened 2023-03-01
DROID-152: Fix initialization issue with prediction linearization

jjohnstz:


miwright2 (approved):
Looks good.  Tested with simulated data and fixes the bug.

PR #60, opened 2023-03-14
DROID-158 Bug Fix

miwright2:
- Not refreshing session information.


angrygorilla (approved):
Tested Framework changes on pixel 5 and pixel 7 using current develop of production app. For each probe power off/power on I saw a new session ID in the exported CSV. Nice catch!

PR #69, opened 2023-04-13
DROID-179: MeatNet Log Transfer

miwright2:
- MeatNet log transfer messages.
- Add in No Route connection state.
- Add in ping for checking when there isn't a route through the network.
- Arbitrate connection state, fw ver, hw ver, sku/lot.
- Fix issues when reading fw ver, hw rev, sku/lot from repeaters.

Not required, but should probably use the corresponding prod app branch.

### DIFF
--- a/combustion-android-ble/src/main/java/inc/combustion/framework/ble/ProbeManager.kt
+++ b/combustion-android-ble/src/main/java/inc/combustion/framework/ble/ProbeManager.kt
@@ -71,6 +71,8 @@ internal class ProbeManager(
         private const val PROBE_STATUS_NOTIFICATIONS_IDLE_TIMEOUT_MS = 15000L
         private const val PROBE_STATUS_NOTIFICATIONS_POLL_DELAY_MS = 30000L
         private const val PROBE_INSTANT_READ_IDLE_TIMEOUT_MS = 5000L
+        private const val IGNORE_PROBES = true
+        private const val IGNORE_REPEATERS = false
     }
 
     // encapsulates logic for managing network data links
@@ -151,6 +153,9 @@ internal class ProbeManager(
             return _probe.value.maxSequenceNumber
         }
 
+    // signals when logs are no longer being added to LogManager.
+    var logTransferCompleteCallback: () -> Unit = { }
+
     private val predictionManager = PredictionManager()
 
     private val instantReadFilter = InstantReadFilter()
@@ -166,6 +171,8 @@ internal class ProbeManager(
     fun addJob(job: Job) = jobManager.addJob(job)
 
     fun addProbe(probe: ProbeBleDevice, baseDevice: DeviceInformationBleDevice, advertisement: CombustionAdvertisingData) {
+        if(IGNORE_PROBES) return
+
         if(arbitrator.addProbe(probe, baseDevice)) {
             handleAdvertisingPackets(probe, advertisement)
             observe(probe)
@@ -174,6 +181,8 @@ internal class ProbeManager(
     }
 
     fun addRepeatedProbe(repeatedProbe: RepeatedProbeBleDevice, repeater: DeviceInformationBleDevice, advertisement: CombustionAdvertisingData) {
+        if(IGNORE_REPEATERS) return
+
         if(arbitrator.addRepeatedProbe(repeatedProbe, repeater))  {
             handleAdvertisingPackets(repeatedProbe, advertisement)
             observe(repeatedProbe)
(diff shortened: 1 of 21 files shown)


combustion-android-ble/src/main/java/inc/combustion/framework/ble/ProbeManager.kt
```diff
@@ -314,33 +388,42 @@ internal class ProbeManager(
...
-            // if using meatnet, then we use the arbitrated connection state
-            _probe.value = _probe.value.copy(baseDevice = _probe.value.baseDevice.copy(connectionState = meatNetState))
-
+        // if don't have a route to send UART messages and we are currently in an active
+        // Upload State, coordinate with Log Manager that it can no longer expect log
+        // messages, and transition the upload state to Unavailable.  Null SessionInfo
+        // so that it can be requested again when a route is established
+        if(arbitrator.hasNoUartRoute && uploadState != ProbeUploadState.Unavailable) {
+            sessionInfo = null
+            uploadState = ProbeUploadState.Unavailable
```
nick-sasquatch, line 401: It looks like this will event out the probe, and you event again after this `if` block--just want to make sure that you don't expect any side effects of `logTransferCompleteCallback()` to happen before the probe state's evented out.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/uart/meatnet/NodeProbeStatusRequest.kt
```diff
@@ -67,7 +62,7 @@ internal class NodeProbeStatusRequest(
             val serialNumber: UInt = data.getLittleEndianUInt32At(SERIAL_NUMBER_INDEX)
             val hopCount: HopCount = HopCount.fromUByte(data[HOP_COUNT_INDEX])
 
-            return NodeProbeStatusRequest(requestId, payloadLength, serialNumber, hopCount, probeStatus)
+            return NodeProbeStatusRequest(requestId, payloadLength, Integer.toHexString(serialNumber.toInt()).uppercase(), hopCount, probeStatus)
```
nick-sasquatch, line 65: There're a few of these conversions around--does it make sense to `typealias` the serial number and create extension methods for it? No's an ok answer.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/device/RepeatedProbeBleDevice.kt
```diff
@@ -156,36 +180,54 @@ internal class RepeatedProbeBleDevice (
     override suspend fun readModelInformation() = uart.readModelInformation()
 
     suspend fun readProbeFirmwareVersion() {
-        Log.d(LOG_TAG, "MeatNet: readProbeFirmwareVersion")
+        val channel = Channel<Unit>(0)
```
nick-sasquatch, line 183: I can't comment on lines that weren't modified, but the `override`ed calls immediately above this don't call `routeMonitor.activity()`--should they?

combustion-android-ble/src/main/java/inc/combustion/framework/ble/device/RepeatedProbeBleDevice.kt
```diff
@@ -156,36 +180,54 @@ internal class RepeatedProbeBleDevice (
...
-                uart.hardwareRevision = resp.hardwareRevision
+                _deviceInfoHardwareRevision = resp.hardwareRevision
+                Log.d(LOG_TAG, "MeatNet: readProbeHardwareRevision: $_deviceInfoHardwareRevision")
+            }
+            uart.owner.lifecycleScope.launch {
+                channel.send(Unit)
             }
         }
         sendUartRequest(NodeReadHardwareRevisionRequest(probeSerialNumber))
+        channel.receive()
```
nick-sasquatch, line 213: This is pretty slick.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/device/RepeatedProbeBleDevice.kt
```diff
@@ -230,39 +276,135 @@ internal class RepeatedProbeBleDevice (
...
+                            probeModelInfoHandler.handled(message.success, message)
+                        }
+                    }
 
                     /// Async Requests that are Broadcast on certain events from a Node
-                    is NodeProbeStatusRequest -> handleProbeStatusRequest(message)
-                    is NodeReadSessionInfoResponse -> sessionInfoHandler.handled(message.success, message.sessionInformation)
+                    is NodeProbeStatusRequest -> {
+                        if(message.serialNumber == probeSerialNumber) {
+                            handleProbeStatusRequest(message)
```
nick-sasquatch, line 345: Reading the description of `routeMonitor` above, it only calls out session info as the exception, but it sounds like it's the async events are really the differentiators?

nick-sasquatch (approved):
Looks good--just a couple comments that probably won't warrant action. Nice work!

combustion-android-ble/src/main/java/inc/combustion/framework/ble/device/RepeatedProbeBleDevice.kt
miwright2: Good question -- no.  They are just doing a device information service read, so not using the UART.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/device/RepeatedProbeBleDevice.kt
miwright2: I agree too

combustion-android-ble/src/main/java/inc/combustion/framework/ble/ProbeManager.kt
miwright2: let me check

miwright2: At this point no -- but it seems cleaner to make the callback before in case that assumption isn't true in the future.

PR #70, opened 2023-04-28
DROID-154: Address Crashlytics ANR

miwright2:
- Buffer and change overflow policy for advertisements.


nick-sasquatch (approved):
Looks really clean--nice work.

PR #71, opened 2023-04-28
DROID-200: Address Crashlytics

miwright2:
- In some cases onDestroy of service is called before NetworkManager is initialized.
- Handle that exception when cleaning up NetworkManager in onDestroy.


angrygorilla (approved):
good catch! 🎸

PR #72, opened 2023-04-28
DROID-201: Address Crashlytics

miwright2:
- Handle situations where the service isn't fully started when making a call to stop the service since start API is asynchronous.


angrygorilla (approved):
boom! 💥

PR #75, opened 2023-05-20
DROID-193: Add fields to Probe data class

miwright2:
- Add isPredicting field.
- Add predictionStale field.


angrygorilla (approved):
Great changes! Thanks for adding those new attributes.

PR #76, opened 2023-06-23
DROID-215: Use snapshot kable build

nick-sasquatch:
- See related PR in example app.  https://github.com/combustion-inc/combustion-android-example/pull/34


miwright2 (approved):
I used this code with a 4 hour brisket cook, and 15 minute tenderloin.  Looks good.  I haven't seen any issues.

PR #79, opened 2023-07-01
DROID-222: Handle Null Service Notification

miwright2:
- Bug generated an ANR -- since null is passed in for the notification, we can't start the service as foreground.

Can test using the example app and using the boolean flag in `MainActivity.kt` that disables foreground service.


nick-sasquatch (approved):
Looks good to me.

PR #81, opened 2023-07-10
DROID-224: Display Advertised Temperatures

miwright2:
- Fix issue with arbitrated connection state.  
- directLink only returns a reference when connected, we need to use probeBleDevice instead to arbitrate.


nick-sasquatch (commented):
Looks good to me.

nick-sasquatch (approved):

PR #84, opened 2023-08-02
DROID-235: New Sessions are not being added on probe reconnection through a Node/Repeater

angrygorilla:
- added new check of connectionState in handleConenctionState to reset sessionInfo when not DeviceConnectionState.CONNECTED

Issue was more prevalent if you initially take a probe from a booster, allow it connect and become repeated through the booster, place the probe back in the booster, allow disconnect of booster to complete, then finally remove the probe from the booster and allow it become repeated through the booster again. The new session information would not get updated for that probe so it would use the older sequence min/max and also session id value.

### DIFF
--- a/combustion-android-ble/src/main/java/inc/combustion/framework/ble/ProbeManager.kt
+++ b/combustion-android-ble/src/main/java/inc/combustion/framework/ble/ProbeManager.kt
@@ -501,8 +501,14 @@ internal class ProbeManager(
                 arbitrator.directLinkDiscoverTimestamp = null
             }
 
-            if(device == logTransferLink) {
-                finishLogTransfer()
+            if(connectionState != DeviceConnectionState.CONNECTED) {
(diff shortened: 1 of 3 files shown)


combustion-android-ble/src/main/java/inc/combustion/framework/ble/ProbeManager.kt
```diff
@@ -501,8 +501,14 @@ internal class ProbeManager(
                 arbitrator.directLinkDiscoverTimestamp = null
             }
 
-            if(device == logTransferLink) {
-                finishLogTransfer()
+            if(connectionState != DeviceConnectionState.CONNECTED) {
+                Log.d(LOG_TAG, "PM($serialNumber): ${device.productType}[${device.id}] No longer connected!  Clearing session info!")
```
miwright2, line 505: Have this be a `Log.i` like the ones above and below.

miwright2 (approved):
Looks good to me!

PR #85, opened 2023-08-04
DROID-233: Kable 0.25.1

miwright2:
- Update to Kable 0.25.1
- Resolves crash from Kable resolved here: https://github.com/JuulLabs/kable/issues/540
- Incorporates `isConnectable` property added to Kable here:  https://github.com/JuulLabs/kable/issues/210


nick-sasquatch (approved):
🔥

PR #87, opened 2023-08-10
DROID-117: Add ability to only have one DFU running at a time

angrygorilla:
- updated DfuBleDevice to have a enabled/disabled state which is reflected via DfuState
- update DfuManager to have a attribute that reflects if a DFU has been started
- added completion callback that is called when a DFU ends with Success, Error, or Aborted; All other devices get re-enabled


miwright2 (approved):
Looks good to me.  @angrygorilla demo'ed the code and walked through with me.  Build full app with related prod app PR.

PR #89, opened 2023-08-17
DROID-238, DROID-241, DROID-242, DROID-243, DROID-236

miwright2:
- DROID-241, DROID-243: Refactor how missing records are determined.
- DROID-241, DROID-243: Transition to upload complete when 2 device status messages have been received w/o a log response. 
- DROID-238: Setup node message completion handling based on request id.
- DROID-238: 30 second session info response indicates no route.
- DROID-238: Send Requests to all nodes and handle first response (excluding log requests)
- DROID-238: Remove Monitor MeatNet coroutine (i.e. SessionInfo ping).
- DROID-238: Accept ProbeStatus messages over any MeatNet node and update arbitration logic.
- DROID-238, DROID-242: Remove message handling based on serial number.
- DROID-238, DROID-242: Request all device information (until response is received) and session info on every probe status message.
- DROID-236: Add Log Upload Percent
- DROID-236: Add expected number of new logs to UploadProgress

### DIFF
--- a/combustion-android-ble/src/main/java/inc/combustion/framework/ble/device/ProbeBleDevice.kt
+++ b/combustion-android-ble/src/main/java/inc/combustion/framework/ble/device/ProbeBleDevice.kt
@@ -120,7 +120,7 @@ internal class ProbeBleDevice (
     override fun connect() = uart.connect()
     override fun disconnect() = uart.disconnect()
 
-    override fun sendSessionInformationRequest(callback: ((Boolean, Any?) -> Unit)?)  {
+    override fun sendSessionInformationRequest(reqId: UInt?, callback: ((Boolean, Any?) -> Unit)?)  {
         sessionInfoHandler.wait(uart.owner, PROBE_MESSAGE_RESPONSE_TIMEOUT_MS, null, callback)
         sendUartRequest(SessionInfoRequest())
     }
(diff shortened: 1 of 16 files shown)


combustion-android-ble/src/main/java/inc/combustion/framework/ble/device/ProbeBleDevice.kt
```diff
@@ -146,9 +146,39 @@ internal class ProbeBleDevice (
     }
 
     override suspend fun readSerialNumber() = uart.readSerialNumber()
+
     override suspend fun readFirmwareVersion() = uart.readFirmwareVersion()
+    fun readProbeFirmwareVersion(callback: (FirmwareVersion) -> Unit) {
+        uart.owner.lifecycleScope.launch {
+            readFirmwareVersion()
+        }.invokeOnCompletion {
+            deviceInfoFirmwareVersion?.let {
```
nick-sasquatch, line 155: Should we be calling back with a default/invalid value if `deviceInfoFirmwareVersion` is `null`? As-is, the caller won't ever know if the read failed.

Looking through the code it doesn't appear that this is much of an issue, so we can probably rack it up to tech debt.

nick-sasquatch (approved):
Just one small comment--looks great.

PR #90, opened 2023-09-26
DROID-253: Resolve Bug

miwright2:
- Cancel message completion timeout whenever we start a request with a new MeatNet request ID.


angrygorilla (approved):
Tested with latest develop version of android app. Good catch!

PR #93, opened 2023-10-25
DROID-106: Read RSSI while connected

miwright2:
- Re-enable code to poll on RSSI while connected.
- Change poll rate from 2s to 0.5s.

Tested by observing RSSI while connected using the debug mode in the prod app.  I tried repeatedly setting predictions while downloading logs, which is the original cause of the issue. I believe that recently updating the Kable dependency to the latest release resolved this issue for us.


nick-sasquatch (approved):
These changes look good. I agree on your slack comment about potentially needing to tune the `REMOTE_RSSI_POLL_RATE_MS` value, but this seems like a good start.

PR #94, opened 2023-10-27
Feature/droid 269 update to android 14

nick-sasquatch:


combustion-android-ble/src/main/java/inc/combustion/framework/service/CombustionService.kt
```diff
@@ -113,7 +115,11 @@ class CombustionService : LifecycleService() {
 
     private fun startForeground() {
         serviceNotification?.let {
-            startForeground(notificationId, serviceNotification)
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
```
nick-sasquatch, line 118: This seems to be required as of Android 14 (https://developer.android.com/about/versions/14/changes/fgs-types-required).

combustion-android-ble/src/main/AndroidManifest.xml
```diff
@@ -1,11 +1,11 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    xmlns:tools="http://schemas.android.com/tools"
-    package="inc.combustion.framework">
```
angrygorilla, line 4: why did this need to be moved out? just out of curiosity.

angrygorilla (approved):
nice job!

angrygorilla:
Tested out build with corresponding app PR on pixel 8 android 14. I will test on older pixel 5 with android 11 (can't remember phone is dead)

combustion-android-ble/src/main/AndroidManifest.xml
nick-sasquatch: I think that was a miss--thanks for catching it!

PR #96, opened 2023-11-16
DROID-278: PoC Producing/Consuming Probe Status

miwright2:
- Make min/max sequence number optional to be consistent with iOS.
- Add a deprecation warning to the existing fields.

https://github.com/combustion-inc/combustion-android-prod/pull/109


nick-sasquatch (approved):
LGTM!

PR #99, opened 2023-11-30
DROID-296: Invalidate probe FW version on return from DFU

nick-sasquatch:
* Fix a problem where new functionality isn't available because the app thinks it's old firmware.
* Fix issue where probe product type wasn't set


angrygorilla (approved):
👍🏻

PR #100, opened 2023-11-30
DROID-297: Fix issue where fw update status doesn't trigger in app

nick-sasquatch:


combustion-android-ble/src/main/java/inc/combustion/framework/ble/NetworkManager.kt
```diff
@@ -264,7 +264,6 @@ internal class NetworkManager(
             serialNumber = probe.probeSerialNumber,
             owner = owner,
             settings = settings,
-            dfuConnectedNodeCallback = { },
```
nick-sasquatch, line 267: This, oddly, wasn't used. I'm guessing we never found a way to integrate the logic with the firmware version coming later.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/ProbeManager.kt
```diff
@@ -622,6 +621,7 @@ internal class ProbeManager(
         _probe.value = _probe.value.copy(
             baseDevice = _probe.value.baseDevice.copy(
                 connectionState = connectionState,
+                fwVersion = fwVersion,
```
nick-sasquatch, line 624: It seems like this is the right approach, though if it was omitted in the past, it's not clear why, and the comment reflects the new implementation.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/NetworkManager.kt
angrygorilla: 🤷🏻

angrygorilla (approved):
🔥

PR #101, opened 2023-12-04
DROID-291: Start parsing of food safe status in the correct raw probe…

nick-sasquatch:
… status array index


angrygorilla (approved):
Verified changes with android app on DROID-259!

PR #103, opened 2023-12-12
DROID-230: Reject log responses that have a higher sequence number than the current max

nick-sasquatch:


miwright2 (approved):
Looks good to me

PR #104, opened 2023-12-14
DROID-281: Add EWMA-smoothed RSSI support for probes

nick-sasquatch:
Please take a close eye at the API here--my goals are to:
* expose when devices are encountered (see `DeviceManager.deviceProximityFlow`); and
* make it straightforward to collect smoothed RSSI data (see `DeviceManager.deviceSmoothedRssiFlow`) from the device

This is aimed very specifically at probes in this PR, but the API shouldn't preclude adding support for nodes transparently at a later date.


nick-sasquatch:
I'm going to take a closer look at this implementation in light of the changes made for #109.

PR #105, opened 2023-12-20
Disable Read RSSI while Connected

miwright2:
UART writes would deadlock, and no commands would make it over the network.  This works around that issue and avoids the deadlock.

- Tested 2 direct connected probes.
- Tested with 3 probes and two nodes.  I do see occasional drops on the response over MeatNet, but the value made it to the probe as observed by the setting on the app/timer; so confident that the write made it to MeatNet.


angrygorilla (approved):
Tested with 3 probes, 2 boosters, timer, 2 apps. Things work much better!

PR #106, opened 2023-12-21
DROID-320: Resolve Issues with Incorrect Connection State

miwright2:
- Publish current connection state on each status notification so it doesn't go stale.
- Don't publish NO_ROUTE connection state during log upload, since we expect session info responses to get dropped when there is lots of traffic over MeatNet.


angrygorilla (approved):
connection state for many probes stayed correct during several log transfers! nice work!

PR #108, opened 2023-12-27
Feature/droid 311 explore moving gradle groovy build files to kotlin dsl

nick-sasquatch:


settings.gradle.kts
```diff
@@ -0,0 +1,24 @@
...
+
+pluginManagement {
+    repositories {
+        google()
+        mavenCentral()
+        gradlePluginPortal()
+    }
+}
+
+@Suppress("UnstableApiUsage")
```
miwright2, line 11: What warning is this suppressing?

miwright2 (approved):
Looks good.  I'm able to sync and build the app with the new setup.

settings.gradle.kts
nick-sasquatch: Apparently `repositoriesMode` ([getRepositoriesMode()](https://docs.gradle.org/current/kotlin-dsl/gradle/org.gradle.api.initialization.resolve/-dependency-resolution-management/get-repositories-mode.html)) is marked as incubating. I don't know if the `@Incubating` is about the Kotlin DSL API itself or if the groovy implementation didn't have a way to represent that the API was subject to change.

PR #109, opened 2024-01-04
DROID-319: Use advertising data as source of RSSI instead of direct read

nick-sasquatch:
Tested with multiple devices, confirmed through debug view that RSSI is being updated.


miwright2 (approved):
Looks great.

PR #113, opened 2024-01-24
DROID-282: Implement association/linking

nick-sasquatch:
@angrygorilla please take a look at this as well.

### DIFF
--- a/combustion-android-ble/src/main/java/inc/combustion/framework/ble/DataLinkArbitrator.kt
+++ b/combustion-android-ble/src/main/java/inc/combustion/framework/ble/DataLinkArbitrator.kt
@@ -28,7 +28,8 @@
 
 package inc.combustion.framework.ble
 
-import android.se.omapi.Session
+import android.util.Log
+import inc.combustion.framework.LOG_TAG
 import inc.combustion.framework.ble.device.*
 import inc.combustion.framework.ble.device.ProbeBleDevice
 import inc.combustion.framework.ble.device.ProbeBleDeviceBase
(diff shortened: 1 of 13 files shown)


combustion-android-ble/src/main/java/inc/combustion/framework/ble/DataLinkArbitrator.kt
```diff
@@ -121,15 +122,40 @@ internal class DataLinkArbitrator(
...
+     *   this selection.
+     */
+    fun finish(jobKey: String? = null, deviceIdsToDisconnect: Set<DeviceID>? = null) {
+        Log.d(LOG_TAG, "DataLinkArbitrator.finish($jobKey, $deviceIdsToDisconnect)")
+        networkNodes.values
+            .forEach {
+                it.finish(jobKey, deviceIdsToDisconnect?.contains(it.id) ?: true)
+            }
+
+        directLink?.disconnect()
```
miwright2, line 149: Disconnect here is a little bit hidden.  Not sure if it can be done higher up, but I'm not sure if we take explicit BLE actions otherwise in this class.

miwright2 (approved):
Looks good.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/DataLinkArbitrator.kt
nick-sasquatch: Good call. Addressed in follow-up commit by moving this functionality up to the caller.

PR #115, opened 2024-02-23
DROID-350: Clean up log manager on probe unlink

nick-sasquatch:


miwright2 (approved):

combustion-android-ble/src/main/java/inc/combustion/framework/log/LogManager.kt
```diff
@@ -242,6 +242,11 @@ internal class LogManager {
         }
     }
 
+    fun finish(serialNumber: String) {
```
miwright2, line 245: Good catch.

PR #116, opened 2024-02-27
DROID-352: Framework Scoping

miwright2:
- Change scoping of some utility functions so they are available to applications.


nick-sasquatch:
LGTM!

nick-sasquatch (approved):

PR #119, opened 2024-03-26
DROID-386: Add game integrated category

nick-sasquatch:


angrygorilla (approved):
Tested with corresponding Android PR and confirmed being able to set 'Game' as the protein.

PR #123, opened 2024-06-11
DROID-401: Thermometer card can sometimes show offline incorrectly

nick-sasquatch:
- **DROID-401: Thermometer card shows stale when not**
- **Fix issue with prediction status not being set correctly**

### DIFF
--- a/combustion-android-ble/src/main/java/inc/combustion/framework/ble/PredictionManager.kt
+++ b/combustion-android-ble/src/main/java/inc/combustion/framework/ble/PredictionManager.kt
@@ -34,6 +34,16 @@ import inc.combustion.framework.service.ProbePredictionState
 import inc.combustion.framework.service.ProbePredictionType
 import java.util.*
 
+/**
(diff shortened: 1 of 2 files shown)


angrygorilla:
@nick-sasquatch  For the second issue `Fix issue with prediction status not being set correctly` What was occurring and how would you test this for this change?

nick-sasquatch:
> @nick-sasquatch For the second issue `Fix issue with prediction status not being set correctly` What was occurring and how would you test this for this change?

IIRC, it was an issue introduced by the first commit, so it's not fixing an existing issue.

Without digging too much into the weeds, I was updating the prediction manager (and thus the linearization), but I wasn't copying the updated linearization value into the local probe instance on normal mode updates. What you'd see is the prediction flicker into a different status on every normal mode packet.

nick-sasquatch:
Closing this in favor of #124 , as it includes these changes and a fix to an issue in here as well.

PR #124, opened 2024-06-13
DROID-412: Add linearization processing to cloud data

nick-sasquatch:
Note that this is built on DROID-401.


nick-sasquatch:
This PR also addresses DROID-401 (previously in PR #123).

angrygorilla (approved):
Good stuff! Tested with Dev board playback, production probe, Display. Tested long cooks and short cooks. Probe connect and disconnect scenarios!

PR #127, opened 2024-06-28
DROID-417: MeatNet node scanning timeouts

nick-sasquatch:
Also fix an issue with multiple MeatNet nodes.

No public API changes.


miwright2 (approved):
These changes make sense to me.  It looks like the bugs were introduced as part of [this refactor](https://github.com/combustion-inc/combustion-android-ble/commit/ab2420ddcfd990ac076a664aafb09e2070f566c9).

PR #130, opened 2024-10-01
Removed cap on prediction time to support low-and-slow.

jmaha:
See title. This bug was reported by several customers on Reddit.


angrygorilla (approved):
Good catch!

PR #133, opened 2024-11-13
Fix bug with request handling

benjcleveland:
The genericRequestHander wasn't tracking the requestId. This caused it to return responses for incorrect messages if other responses were interleaved. Passing the requestId to the wait and handle methods fixes this issue.


miwright2 (approved):
Looks good.

PR #137, opened 2024-12-05
Increase wifi timeouts and make sendNodeRequest call threadsafe

jhavatar:
https://linear.app/combustion/issue/DROID-530/wifi-connect-times-out
https://linear.app/combustion/issue/DROID-531/synchronize-wifi-calls-to-ble-device

## Description
- Wifi related calls are timing out
- `NodeBleDevice..sendNodeRequest()` calls are not threadsafe - observed calls getting stuck callback never happening

**Note, first commit is easier to read since does not contain autoformatting**

## Tech Notes
- increase timeout to 30000millis (30s) since WIFI_CONNECT call still timeouts at 15s and fast calls will finish fast anyway
- `Semaphore`s used in `NetworkManager` to ensure a single node request at a time per node.

### DIFF
--- a/combustion-android-ble/src/main/java/inc/combustion/framework/ble/device/NodeBleDevice.kt
+++ b/combustion-android-ble/src/main/java/inc/combustion/framework/ble/device/NodeBleDevice.kt
@@ -7,8 +7,12 @@ import androidx.lifecycle.lifecycleScope
 import inc.combustion.framework.LOG_TAG
 import inc.combustion.framework.ble.NetworkManager
 import inc.combustion.framework.ble.scanning.CombustionAdvertisingData
-import inc.combustion.framework.ble.uart.meatnet.*
+import inc.combustion.framework.ble.uart.meatnet.GenericNodeRequest
+import inc.combustion.framework.ble.uart.meatnet.GenericNodeResponse
 import inc.combustion.framework.ble.uart.meatnet.NodeReadFeatureFlagsRequest
+import inc.combustion.framework.ble.uart.meatnet.NodeReadFeatureFlagsResponse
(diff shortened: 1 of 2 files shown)


combustion-android-ble/src/main/java/inc/combustion/framework/ble/device/NodeBleDevice.kt
```diff
@@ -23,34 +27,62 @@ internal class NodeBleDevice(
...
+            return uart.isConnectable.get()
+        }
 
-    val deviceInfoSerialNumber: String? get() { return uart.serialNumber }
+    val deviceInfoSerialNumber: String?
+        get() {
+            return uart.serialNumber
+        }
 
+    @Synchronized
```
benjcleveland, line 71: What was the problem was happening when this was called by different threads?

The call to `genericRequestHandler.wait` sets variables that are reused/reset with the next call to `sendNodeRequest`. Does this prevent that from happening?

jhavatar: Ah, you have a point - I made some other change and that actually fixed the issue and I just assumed... Let me see if there is an easy equivalent fix or I will just remove.

jhavatar: The problem is that calls get stuck with no response callbacks happening.

jhavatar: Switched to a `Semaphore` solution in `NetworkManager`.

benjcleveland (approved):

PR #145, opened 2025-01-13
DROID-562 : Fix node firmware state not being observed

jhavatar:
## Description
- There is a requirement to display a new message when node update is critical but app's device list screen was not observing any node firmware states
- Add missing/removed publishing of node's firmware state

**Note, required for app PR https://github.com/combustion-inc/combustion-android-prod/pull/292**

## Video
Critical update message is displayed and wifi device is still detected
https://github.com/user-attachments/assets/fda6aae8-ce6a-405c-bbe8-ae49108e44a2


nick-sasquatch (approved):
Looks good. Thanks for picking it up.

PR #147, opened 2025-02-07
Power Mode and Probe reset funcitonality

jhavatar:
https://linear.app/combustion/issue/DROID-558/add-probe-powermode-to-probestatus
https://linear.app/combustion/issue/DROID-557/add-probe-power-mode-update-requestresponse
https://linear.app/combustion/issue/DROID-586/add-reset-session-requestresponse-ble-library

## Tech Notes


angrygorilla (approved):
Looks good!

PR #148, opened 2025-02-17
DROID-595 : rename power mode related classes to match cloud requirements

jhavatar:
https://linear.app/combustion/issue/DROID-595/write-power-mode-value-to-v1probe-status-and-read-from-firestore

**Note, required by PR https://github.com/combustion-inc/combustion-android-prod/pull/304**

## Description
Rename power mode related classes to match cloud requirements


nick-sasquatch (approved):
LGTM

PR #149, opened 2025-02-18
DROID-590 : logic used in app's samsung fix

jhavatar:
https://linear.app/combustion/issue/DROID-590/graph-rarely-appearing-on-samsung-phones

**Note, required for PR https://github.com/combustion-inc/combustion-android-prod/pull/305**

## Description
- add functionality to `StateFlowMutableMap` that is used in app to fix one of the samsung issues


nick-sasquatch (approved):
Looks good!

PR #150, opened 2025-02-20
DROID-600 : update android SDK to 35

jhavatar:
**Note, merges to gauge feature branch**

## Description
App-feature requires a library version that requires updating the app to SDK 35 (https://github.com/combustion-inc/combustion-android-prod/pull/307). To support compiling the framework from source in the SDK 35 app:
 - update the framework to SDK 35
 - update dependency libraries with version warnings
 - fix code warnings


angrygorilla (approved):
Nice work!

PR #151, opened 2025-03-03
DROID-608 : Detect stuck dfu and retry after 10s

jhavatar:
https://linear.app/combustion/issue/DROID-603/samsung-phones-bricking-probes-during-ota
https://linear.app/combustion/issue/DROID-608/if-device-is-advertising-dfu-names-for-longer-than-10s-then-auto
https://linear.app/combustion/issue/DROID-609/log-instances-of-either-emergency-recovery-or-normal-dfu-operations-to

## Description
- Detect stuck dfu and retry after 10s
- Framework provides dfu analytics events

### Note
- **can test with app PR https://github.com/combustion-inc/combustion-android-prod/pull/310**
- Could not get naming DFU to match iOS working in this PR - still legacy names
- updated nordic and kable lib versions - failed to fix naming issue but still has value

## Tech Notes
- `DeviceScanner` now also detects bootloading advertisements and emits them in `bootloadingAdvertisements` flow
- `DfuManager` now calls `collectBootloaderDevices()` that observes `DeviceScanner.bootloadingAdvertisements` as `BootLoaderDevice`
- DFU logic form `DfuBleDevice` extracted to new `PerformDfuDelegate` that is used by both `DfuBleDevice` and `BootLoaderDevice`
- `DfuManager` restarts DFU process if `BootLoaderDevice` has been observed for longer than 10s
- `DeviceManager.startCombustionService` now (optionally) takes latest firmware resources as a parameter since required for restarting DFU process
- Create `AnalyticsTracker` that tracks dfu_started an df_retry events from `DfuManager`
- Provide `AnalyticsTracker`'s flow of analytics events to app with `DeviceManager.analyticsFlow`


angrygorilla (approved):
Nice job @jhavatar !!

PR #162, opened 2025-05-30
Fix app v2.5.1 thinks Gauge is an overheating probe

jhavatar:
## Desciption 
- Fix app v2.5.1 (pre gauge support) thinks Gauge is an overheating probe

## Tech Notes
- When handling advertisements ignore those from product types that are not supported, i.e. only consider PROBE and DISPLAY (renamed to NODE in gauge branch)


angrygorilla (approved):
Nice work! No longer see overheating probes!

PR #164, opened 2025-06-05
DROID-638 : fix Unknown device type crash when updating connected nodes

jhavatar:
## Description
- Fix unknown device type crash when updating connected nodes

## Tech Notes
- unlinking gauge, which is a hybrid node device, removed reference to its node component (same mac) which is still "used" by devices


angrygorilla (approved):
Nice work! We found it!

PR #166, opened 2025-06-13
Fix bug: Gauge connection goes stale after ~15 seconds

jhavatar:
## Description
- handle repeated gauge status request
- notify activity when gauge receives status message while it has no sensor.


angrygorilla (approved):
NIce job! Tested out with my pre-prod gauge!

PR #171, opened 2025-09-05
DROID-687: App crash if access framework's service before it was initialized

jhavatar:
## Description
lateinit `DeviceManager.service` wont exist until service is initialized -- will throw exception if accessed before then

## Tech Notes
Implement `doWhenServiceInitialized` that will delay function on service (when service is not initialized) until service has been initialized - e.g. `startDfuMode()`

## Reproduce / Verify Fix
1. In device's Developer options enable "Don't keep activities" and set "Background process limit" to "No background processes"
2. launch Combustion app and navigate to "Update Devices" and move app to background
3. launch other apps
4. return to Combustion app that should still be on functioning "Update Devices" screen and not crashing


angrygorilla (approved):
Nice work! I wasn't able to recreate this issue locally. We should just keep monitoring for regressions for this issue.

PR #172, opened 2025-09-08
DROID-688: Remove Gauge battery percentage

jhavatar:
## Description
Remove Gauge battery percentage

## Tech Notes
Simplify detecting low battery for device


angrygorilla (approved):
Checked changes with gauge dev board and other devices!

PR #182, opened 2026-02-22
[PRO-132] Add high radio power bit to ThermometerPreferences

jmaha:
## Summary
- Adds `highRadioPower: Boolean` property to `ThermometerPreferences` data class, parsed from bit 2 of the preferences byte
- Updates `fromUByte()` to extract the new bit and `DEFAULT` to include `highRadioPower = false`
- No changes needed to `ProbePowerMode.kt` — the `0x03` mask correctly isolates bits 0-1

## Context
Corresponds to firmware PRO-132 which adds a high radio power status bit (bit 2) to the ThermometerPreferences BLE advertising byte. The Android BLE library exposes smoothed RSSI via flows without internal thresholding, so proximity threshold adjustments for high-power devices should be handled at the app layer using the new `highRadioPower` flag.

## Test plan
- [ ] Build Android BLE library successfully
- [ ] Run unit tests
- [ ] Connect to V2 probe with high power enabled — verify `highRadioPower` parses as `true`
- [ ] Connect to V1 probe — verify `highRadioPower` defaults to `false`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/combustion-android-ble/src/main/java/inc/combustion/framework/service/ThermometerPreferences.kt
+++ b/combustion-android-ble/src/main/java/inc/combustion/framework/service/ThermometerPreferences.kt
@@ -28,15 +28,22 @@
 
 package inc.combustion.framework.service
 
-data class ThermometerPreferences(val powerMode: ProbePowerMode) {
+data class ThermometerPreferences(
+    val powerMode: ProbePowerMode,
+    val highRadioPower: Boolean = false,
+) {
 
     companion object {
         const val PROBE_PREFS_SIZE_BYTES = 1
 
-        val DEFAULT = ThermometerPreferences(powerMode = ProbePowerMode.NORMAL)
+        val DEFAULT = ThermometerPreferences(powerMode = ProbePowerMode.NORMAL, highRadioPower = false)
 
         fun fromUByte(byte: UByte): ThermometerPreferences {
-            return ThermometerPreferences(powerMode = ProbePowerMode.fromUByte(byte))
+            val highRadioPower = (byte.toInt() shr 2) and 0x1 != 0
+            return ThermometerPreferences(
+                powerMode = ProbePowerMode.fromUByte(byte),
+                highRadioPower = highRadioPower,
+            )
         }
     }
 }


jhavatar (approved):

combustion-android-ble/src/main/java/inc/combustion/framework/service/ThermometerPreferences.kt
```diff
@@ -28,15 +28,22 @@
...
 
     companion object {
         const val PROBE_PREFS_SIZE_BYTES = 1
 
-        val DEFAULT = ThermometerPreferences(powerMode = ProbePowerMode.NORMAL)
+        val DEFAULT = ThermometerPreferences(powerMode = ProbePowerMode.NORMAL, highRadioPower = false)
 
         fun fromUByte(byte: UByte): ThermometerPreferences {
-            return ThermometerPreferences(powerMode = ProbePowerMode.fromUByte(byte))
+            val highRadioPower = (byte.toInt() shr 2) and 0x1 != 0
```
angrygorilla, line 42: Style nit: `shr 2` is a magic number, and the bit-extraction style is inconsistent with the sibling `ProbePowerMode.fromUByte`, which uses `UByte`/`UShort` masking with a named `PROBE_POWER_MODE_MASK` constant.

Consider mirroring that pattern:

```kotlin
private const val HIGH_RADIO_POWER_MASK: UByte = 0x04u
...
val highRadioPower = (byte and HIGH_RADIO_POWER_MASK) != 0u.toUByte()
```

A named constant documents which bit is being read and keeps both parsers in the same idiom.

combustion-android-ble/src/main/java/inc/combustion/framework/service/ThermometerPreferences.kt
```diff
@@ -28,15 +28,22 @@
...
+    val highRadioPower: Boolean = false,
+) {
 
     companion object {
         const val PROBE_PREFS_SIZE_BYTES = 1
 
-        val DEFAULT = ThermometerPreferences(powerMode = ProbePowerMode.NORMAL)
+        val DEFAULT = ThermometerPreferences(powerMode = ProbePowerMode.NORMAL, highRadioPower = false)
 
         fun fromUByte(byte: UByte): ThermometerPreferences {
```
angrygorilla, line 41: No unit test is being added for this parser. `ProbePowerModeTest.kt` exists with parameterized cases on the same byte — a `ThermometerPreferencesTest` alongside it would lock the bit position down and protect against regressions when bits 3+ get assigned later.

Suggested cases:
- `0x00` → `powerMode=NORMAL, highRadioPower=false`
- `0x04` → `powerMode=NORMAL, highRadioPower=true`
- `0x05` → `powerMode=ALWAYS_ON, highRadioPower=true`
- `0x07` → confirm bits beyond 0–2 don't leak in

The PR test plan only lists "Run unit tests" — there's nothing new to run. Worth adding given parsing is the entire surface of this change.

combustion-android-ble/src/main/java/inc/combustion/framework/service/ThermometerPreferences.kt
```diff
@@ -28,15 +28,22 @@
...
+data class ThermometerPreferences(
+    val powerMode: ProbePowerMode,
+    val highRadioPower: Boolean = false,
+) {
 
     companion object {
         const val PROBE_PREFS_SIZE_BYTES = 1
 
-        val DEFAULT = ThermometerPreferences(powerMode = ProbePowerMode.NORMAL)
+        val DEFAULT = ThermometerPreferences(powerMode = ProbePowerMode.NORMAL, highRadioPower = false)
```
angrygorilla, line 39: Related concern at `ProbeManager.kt:533` (out of this diff but worth flagging): the `setPowerMode` fallback constructs a fresh `ThermometerPreferences(powerMode = powerMode)` when `thermometerPrefs` is null, which now silently sets `highRadioPower = false` regardless of what was last observed on the wire.

In practice the null branch is rare (status frames typically arrive before any user-initiated power-mode change), but it'll silently regress every future field added here. Suggest changing that fallback to `ThermometerPreferences.DEFAULT.copy(powerMode = powerMode)` — same behavior today, future-proof for any new fields that get a non-default observed value.

angrygorilla (commented):
Inline comments on three items — overall safe to merge, these are robustness/style improvements.

angrygorilla (approved):
Left some feedback items that were minor but up to you on severity. Nice work!

jmaha:
Canceling this vibe-coded PR as it's really not complete as it only covers the Probe and not all the other devices, and related threshold logic.

jmaha:
Sorry for the confusion with this one!

PR #183, opened 2026-03-11
DROID-746 : changes to setting ProbeID requirements

jhavatar:
## Description
Implemented requirement changes:
1. only resolve observed conflicts
2. resolve conflict by assigning lowest ID to highest serialNr
3. If there is a conflict but no free IDs — do nothing
4. Don’t attempt to assign an ID to a device if ID is known to be used by a  different device (manufacture a collision)

### DIFF
--- a/combustion-android-ble/src/main/java/inc/combustion/framework/ble/NetworkManager.kt
+++ b/combustion-android-ble/src/main/java/inc/combustion/framework/ble/NetworkManager.kt
@@ -315,6 +315,8 @@ internal class NetworkManager(
     val silenceAlarmsRequestFlow: SharedFlow<SilenceAlarmsRequest>
         get() = _silenceAlarmsRequestFlow.asSharedFlow()
 
+    val availableProbeIDs: Flow<List<ProbeID>> = probeIdManager.availableProbeIds
+
     init {
         require(context.applicationContext === context) {
             "NetworkManager must be given application context"
@@ -505,9 +507,16 @@ internal class NetworkManager(
         completionHandler: (Boolean) -> Unit,
     ) {
         Log.v(LOG_TAG, "setProbeID: assign $probeId to $serialNumber")
-        probeIdManager.checkAndAvoidProbeIdConflictOnSetProbeId(serialNumber, probeId)
-        probeManagers[serialNumber]?.setProbeID(probeId, completionHandler) ?: run {
+        if (probeIdManager.hasProbeIdConflict(serialNumber, probeId)) {
+            Log.w(
+                LOG_TAG,
+                "setProbeID: unable to perform since there is an existing probe assigned to $probeId",
+            )
             completionHandler(false)
+        } else {
+            probeManagers[serialNumber]?.setProbeID(probeId, completionHandler) ?: run {
+                completionHandler(false)
+            }
         }
     }
 
--- a/combustion-android-ble/src/main/java/inc/combustion/framework/ble/NetworkManager.kt
+++ b/combustion-android-ble/src/main/java/inc/combustion/framework/ble/NetworkManager.kt
@@ -507,6 +507,10 @@ internal class NetworkManager(
         completionHandler: (Boolean) -> Unit,
     ) {
         scope.launch(Dispatchers.Main) {
+            // Serializes concurrent explicit setProbeID calls against each other's conflict check.
+            // Does NOT protect against the gap between dispatching the BLE write and the
+            // observation flow updating knownProbeIdAssignedToDevice; any collision that slips
+            // through that window is caught and resolved by ProbeIdManager.addDevice.
             setProbeIdLock.withLock {
                 Log.v(LOG_TAG, "setProbeID: assign $probeId to $serialNumber")
                 if (probeIdManager.hasProbeIdConflict(serialNumber, probeId)) {
--- a/combustion-android-ble/src/main/java/inc/combustion/framework/ble/NetworkManager.kt
+++ b/combustion-android-ble/src/main/java/inc/combustion/framework/ble/NetworkManager.kt
@@ -53,15 +53,12 @@ import inc.combustion.framework.service.CombustionProductType.NODE
 import inc.combustion.framework.service.CombustionProductType.PROBE
 import inc.combustion.framework.service.utils.ConcurrentSnapshotMap
 import inc.combustion.framework.service.utils.plus
-import kotlinx.coroutines.CoroutineName
-import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.*
 import kotlinx.coroutines.channels.BufferOverflow
 import kotlinx.coroutines.flow.*
-import kotlinx.coroutines.launch
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
 import java.util.concurrent.atomic.AtomicBoolean
-import java.util.concurrent.locks.ReentrantLock
-import kotlin.concurrent.withLock
 import kotlin.random.Random
 import kotlin.random.nextUInt
 
@@ -205,7 +202,7 @@ internal class NetworkManager(
     }
 
     private val probeIdManager = ProbeIdManager(::setProbeID, scope)
-    private val setProbeIdLock = ReentrantLock()
+    private val setProbeIdLock = Mutex()
 
     var deviceAllowlist: Set<String>? = settings.probeAllowlist
         private set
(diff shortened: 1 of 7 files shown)


combustion-android-ble/src/main/java/inc/combustion/framework/ble/ProbeIdManager.kt
```diff
@@ -43,10 +42,15 @@ private const val LOG_TAG = "ProbeIdManager"
...
+    private val knownProbeIdAssignedToDevice: StateFlowMutableMap<ProbeID, String> = StateFlowMutableMap(),
 ) {
     private val probeIdObservations = ConcurrentHashMap<String, Job>()
 
+    val availableProbeIds: Flow<List<ProbeID>> =
+        knownProbeIdAssignedToDevice.stateFlow.map { map ->
+            ProbeID.entries.filterNot { it in map }
+        }
+
     private fun findLowestAvailableProbeId(exceptConflictId: ProbeID): ProbeID? {
```
angrygorilla, line 54: `findLowestAvailableProbeId` appears to be unused now — it was only called from the deleted `avoidProbeIdConflict`. Should be removed.

angrygorilla (commented):
Overall clean simplification — removing the auto-reassignment logic eliminates a class of subtle race conditions. A few items noted inline.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/ProbeIdManager.kt
```diff
@@ -43,10 +42,15 @@ private const val LOG_TAG = "ProbeIdManager"
 internal class ProbeIdManager(
     private val setProbeID: (serialNumber: String, probeId: ProbeID, completionHandler: (Boolean) -> Unit) -> Unit,
     private val scope: CoroutineScope,
-    private val knownProbeIdAssignedToDevice: ConcurrentHashMap<ProbeID, String> = ConcurrentHashMap<ProbeID, String>(),
+    private val knownProbeIdAssignedToDevice: StateFlowMutableMap<ProbeID, String> = StateFlowMutableMap(),
 ) {
     private val probeIdObservations = ConcurrentHashMap<String, Job>()
 
+    val availableProbeIds: Flow<List<ProbeID>> =
```
angrygorilla: Minor naming inconsistency: this is `availableProbeIds` (camelCase "Ids") while `NetworkManager` and `DeviceManager` use `availableProbeIDs` (uppercase "IDs"). The rest of the codebase uses `ProbeID`, so this should probably be `availableProbeIDs` for consistency.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/NetworkManager.kt
```diff
@@ -505,9 +507,16 @@ internal class NetworkManager(
         completionHandler: (Boolean) -> Unit,
     ) {
         Log.v(LOG_TAG, "setProbeID: assign $probeId to $serialNumber")
-        probeIdManager.checkAndAvoidProbeIdConflictOnSetProbeId(serialNumber, probeId)
-        probeManagers[serialNumber]?.setProbeID(probeId, completionHandler) ?: run {
+        if (probeIdManager.hasProbeIdConflict(serialNumber, probeId)) {
```
angrygorilla: Nit: the check-then-act (`hasProbeIdConflict` then `setProbeID`) is not atomic. Two concurrent `setProbeID` calls for the same `probeId` from different serial numbers could both pass the check. Likely fine given single-threaded UI dispatch, but worth keeping in mind.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/NetworkManager.kt
```diff
@@ -505,9 +507,16 @@ internal class NetworkManager(
...
+            Log.w(
+                LOG_TAG,
+                "setProbeID: unable to perform since there is an existing probe assigned to $probeId",
+            )
             completionHandler(false)
+        } else {
+            probeManagers[serialNumber]?.setProbeID(probeId, completionHandler) ?: run {
```
angrygorilla: Previously `checkAndAvoidProbeIdConflictOnSetProbeId` called `removeProbeIdAssignmentsToDevice` to clean up stale assignments before reassignment. That cleanup no longer happens in the `setProbeID` path — it relies on `addDevice`'s flow observation (ProbeIdManager line 147) to handle it. Confirm that the timing is acceptable: a device could briefly hold two probe ID entries if the flow hasn't emitted yet.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/ProbeIdManager.kt
jhavatar: still used in `resolveProbeIdConflict()`

combustion-android-ble/src/main/java/inc/combustion/framework/ble/ProbeIdManager.kt
jhavatar: updated

combustion-android-ble/src/main/java/inc/combustion/framework/ble/NetworkManager.kt
jhavatar: yes, this matches new expectation of "only resolve observed conflicts"

combustion-android-ble/src/main/java/inc/combustion/framework/ble/NetworkManager.kt
jhavatar: converted to atomic

combustion-android-ble/src/main/java/inc/combustion/framework/ble/NetworkManager.kt
```diff
@@ -504,10 +506,20 @@ internal class NetworkManager(
         probeId: ProbeID,
         completionHandler: (Boolean) -> Unit,
     ) {
-        Log.v(LOG_TAG, "setProbeID: assign $probeId to $serialNumber")
-        probeIdManager.checkAndAvoidProbeIdConflictOnSetProbeId(serialNumber, probeId)
-        probeManagers[serialNumber]?.setProbeID(probeId, completionHandler) ?: run {
-            completionHandler(false)
+        scope.launch(Dispatchers.Main) {
```
angrygorilla, line 509: Behavior change worth flagging: `setProbeID` used to kick off the BLE write synchronously and the caller's `completionHandler` fired when the device responded. After this change the work is queued on `scope.launch(Dispatchers.Main) { setProbeIdLock.withLock { ... } }` — so under contention `completionHandler` may now fire much later than before because it's queued behind another in-flight `setProbeID`. Worth confirming the call sites in `ProbeIdManager.resolveProbeIdConflict` (lines 104, 116) and `DeviceManager.setProbeID` tolerate this.

Also: the lock's actual atomicity is narrower than "check-and-set" implies. `withLock` is held only across `hasProbeIdConflict` + dispatching `probeManagers[serialNumber]?.setProbeID(...)` — the BLE write itself completes asynchronously *after* the lock is released, and `knownProbeIdAssignedToDevice` only updates later via the device-observation flow in `addDevice`. Two near-simultaneous calls for different serials targeting the same `probeId` can both pass the conflict check.

A short code comment near the `withLock` describing what the lock does and does not protect would help future readers avoid relying on guarantees that aren't there.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/ProbeIdManager.kt
```diff
@@ -43,10 +42,15 @@ private const val LOG_TAG = "ProbeIdManager"
 internal class ProbeIdManager(
     private val setProbeID: (serialNumber: String, probeId: ProbeID, completionHandler: (Boolean) -> Unit) -> Unit,
     private val scope: CoroutineScope,
-    private val knownProbeIdAssignedToDevice: ConcurrentHashMap<ProbeID, String> = ConcurrentHashMap<ProbeID, String>(),
+    private val knownProbeIdAssignedToDevice: StateFlowMutableMap<ProbeID, String> = StateFlowMutableMap(),
 ) {
     private val probeIdObservations = ConcurrentHashMap<String, Job>()
 
+    val availableProbeIDs: Flow<List<ProbeID>> =
```
angrygorilla, line 49: Subscribers to this flow (and `DeviceManager.availableProbeIDs` downstream) will see an emission storm any time `removeProbeIdAssignmentsToDevice` runs (line 60 in the unchanged region of this file). That method does `ProbeID.entries.forEach { knownProbeIdAssignedToDevice.remove(it) }` — each `remove` is a separate update on `StateFlowMutableMap`, so a single device removal can trigger up to 7 intermediate emissions of partially-cleaned state.

`StateFlowMutableMap.removeIf` (defined at `StateFlowMutableMap.kt:156`) collapses to one update. Suggest:

```kotlin
knownProbeIdAssignedToDevice.removeIf { it.key != except && it.value == serialNumber }
```

UI consumers of `availableProbeIDs` then see one transition per device removal instead of N.

angrygorilla (commented):
Two inline items on the lock semantics and the emission behavior of the new `availableProbeIDs` flow.

angrygorilla (approved):
Lets some feedback items that were mostly nice to have or potential changes. You can make the call.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/NetworkManager.kt
jhavatar: added comment

combustion-android-ble/src/main/java/inc/combustion/framework/ble/ProbeIdManager.kt
jhavatar: made suggested change

PR #184, opened 2026-04-20
Parse high radio power bit for gauges and other fixes/improvements

jhavatar:
## Description
1. [DROID-751](https://linear.app/combustion/issue/DROID-751/parse-high-radio-power-bit-for-devices):`isHighRadioPower` now available in Gauge and Probe's advertisement -- parse it. 
2. DROID-751:`SpecializedDevice.highRadioPower` interface implemented for Probe, Gauge, Engine
3. ~[DROID-763 ](https://linear.app/combustion/issue/DROID-763/engine-control-device-disconnects): Debounce Engine control-device disconnect confirmation by 30s to avoid flapping the UI on brief radio drops, with tests and thread-safety for concurrent status updates.~ Moved to app to also process cloud data
4. Fix status arbitration to stop a rejected stale status from lowering the bar for later updates.
5. Fix/remove some previous incorrect Engine assumptions, e.g. detach controlSerialNumber and controlDeviceType from controlDeviceConnected
6. Fix concurrent status-handling races across Engine/Gauge/Probe managers with per-manager mutexes while preserve completionHandler's original Main-thread delivery


combustion-android-ble/src/main/java/inc/combustion/framework/ble/scanning/GaugeAdvertisingData.kt
```diff
@@ -83,6 +82,14 @@ internal class GaugeAdvertisingData(
                 manufacturerData.copyOf().sliceArray(HIGH_LOW_ALARM_RANGE)
             )
 
+            val preferences = if (manufacturerData.size > PREFERENCES_RANGE.last) {
+                GaugePreferences.fromRawByte(
+                    manufacturerData.copyOf().sliceArray(PREFERENCES_RANGE)[0]
+                )
+            } else {
+                GaugePreferences.DEFAULT
```
angrygorilla: **Bug**: this should be `null`, not `GaugePreferences.DEFAULT`.

`GaugeManager.updateDataFromAdvertisement` (line 242) writes:
```kotlin
gaugePrefs = advertisement.gaugePreferences ?: updatedGauge.gaugePrefs,
```
The `?: updatedGauge.gaugePrefs` branch is dead code as written — `gaugePreferences` is never null from `create()`. So a short advertisement (legacy firmware, truncated packet) silently clobbers a previously-observed `gaugePrefs` with `DEFAULT` (i.e. resets `highRadioPower` to `false`), even though the BLE manager was clearly written intending to preserve the prior value.

The probe side gets this right: `ProbeAdvertisingData.kt:117` returns `null` for short data, and `ProbeManager.kt:1093-1097` uses `?.let { copy(...) }` to preserve. Mirror that here:

```kotlin
val preferences = if (manufacturerData.size > PREFERENCES_RANGE.last) {
    GaugePreferences.fromRawByte(
        manufacturerData.copyOf().sliceArray(PREFERENCES_RANGE)[0]
    )
} else {
    null
}
```
Then the existing `?: updatedGauge.gaugePrefs` in `GaugeManager` works as intended.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/scanning/ProbeAdvertisingData.kt
```diff
@@ -65,6 +58,7 @@ internal class ProbeAdvertisingData(
         private val MODE_COLOR_ID_RANGE = 18..18
         private val STATUS_RANGE = 19..19
         private val NETWORK_INFO_RANGE = 20..20
+        private val PREFERENCES_RANGE = 22..22
```
angrygorilla, line 69: Why does `PREFERENCES_RANGE` skip byte 21? The other ranges in this file are contiguous:
```
MODE_COLOR_ID_RANGE = 18..18
STATUS_RANGE        = 19..19
NETWORK_INFO_RANGE  = 20..20
PREFERENCES_RANGE   = 22..22   // gap at byte 21
```
Is byte 21 reserved by firmware, or is this an off-by-one? `GaugeAdvertisingData.kt:60` documents its gap explicitly:
```kotlin
private val RESERVED_RANGE = 14..14 // previously BATTERY_PERCENTAGE_RANGE
```
If byte 21 is reserved, please follow the same pattern with a `RESERVED_RANGE = 21..21` constant + comment so future readers don't think this is a typo (and so a future preference bit knows where it can safely land).

combustion-android-ble/src/main/java/inc/combustion/framework/service/GaugePreferences.kt
```diff
@@ -0,0 +1,45 @@
...
+
+data class GaugePreferences(
+    val highRadioPower: Boolean = false,
+) {
+    companion object {
+        val DEFAULT = GaugePreferences(highRadioPower = false)
+
+        fun fromRawByte(byte: UByte): GaugePreferences {
+            return GaugePreferences(
+                highRadioPower = byte.isBitSet(0),
```
angrygorilla: Bit-position spec drift worth documenting in this file (and in `ThermometerPreferences.fromUByte`):

- Gauge: `highRadioPower` is bit **0** of the gauge preferences byte
- Probe: `highRadioPower` is bit **2** of the thermometer preferences byte (because bits 0–1 are `ProbePowerMode`)

This is presumably what the firmware spec dictates, but nothing in either parser documents which BLE/firmware document defines the layout. A one-line comment here referencing the relevant section under `combustion-documentation/` would prevent silent regressions if someone ever tries to "unify" the bit positions across the two devices, or wants to add another preference bit later and has to figure out which positions are taken.

Something like:
```kotlin
// Bit 0 = highRadioPower (per gauge BLE spec §X.Y).
// Note: probes use bit 2 — different layout because bits 0–1 carry ProbePowerMode.
highRadioPower = byte.isBitSet(0),
```

combustion-android-ble/src/main/java/inc/combustion/framework/service/ThermometerPreferences.kt
```diff
@@ -28,15 +28,22 @@
...
 
     companion object {
         const val PROBE_PREFS_SIZE_BYTES = 1
 
-        val DEFAULT = ThermometerPreferences(powerMode = ProbePowerMode.NORMAL)
+        val DEFAULT = ThermometerPreferences(powerMode = ProbePowerMode.NORMAL, highRadioPower = false)
 
         fun fromUByte(byte: UByte): ThermometerPreferences {
-            return ThermometerPreferences(powerMode = ProbePowerMode.fromUByte(byte))
+            val highRadioPower = (byte.toInt() shr 2) and 0x1 != 0
```
angrygorilla: Drive-by cleanup opportunity: this PR adds `GaugePreferences.fromRawByte` which uses the project-standard `byte.isBitSet(0)` extension (matching `GaugeStatusFlags`, `HighLowAlarmStatus`, `OverheatingSensors`, etc.). The `(byte.toInt() shr 2) and 0x1 != 0` style here is now visibly inconsistent next to its sibling.

```kotlin
val highRadioPower = byte.isBitSet(2)
```

Same note I left on PR #182 — easier to land here since the diff already touches both parsers.

combustion-android-ble/src/main/java/inc/combustion/framework/service/GaugePreferences.kt
```diff
@@ -0,0 +1,45 @@
...
+import inc.combustion.framework.isBitSet
+
+data class GaugePreferences(
+    val highRadioPower: Boolean = false,
+) {
+    companion object {
+        val DEFAULT = GaugePreferences(highRadioPower = false)
+
+        fun fromRawByte(byte: UByte): GaugePreferences {
+            return GaugePreferences(
```
angrygorilla, line 44: Nit: method-name asymmetry. `ThermometerPreferences.fromUByte(byte: UByte)` and `GaugePreferences.fromRawByte(byte: UByte)` have the same signature and parallel responsibility but different names.

The codebase is split:
- `fromRawByte`: `GaugeStatusFlags`, `SensorTemperature.fromRawDataStart`, `HighLowAlarmStatus.fromRawData`
- `fromUByte`: `ProbeColor`, `ProbeID`, `ProbeMode`, `ProbeBatteryStatus`, `ProbePowerMode`, `ThermometerPreferences`, `HopCount`

`fromRawByte` is the gauge-side convention and `fromUByte` is the probe-side convention, so this isn't wrong — but worth flagging for the next time someone draws a new convention line. No need to churn this PR.

angrygorilla (commented):
Five inline items: a real bug in the gauge short-advertisement path, a question about a byte gap in probe data, plus three style/consistency items across the two new parsers.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/scanning/ProbeAdvertisingData.kt
jhavatar: 👍 added range OVERHEAT_RANGE = 21..21

combustion-android-ble/src/main/java/inc/combustion/framework/ble/scanning/GaugeAdvertisingData.kt
jhavatar: makes sense

combustion-android-ble/src/main/java/inc/combustion/framework/service/GaugePreferences.kt
jhavatar: going with fromRawByte

combustion-android-ble/src/main/java/inc/combustion/framework/service/GaugePreferences.kt
jhavatar: added comment

combustion-android-ble/src/main/java/inc/combustion/framework/service/ThermometerPreferences.kt
jhavatar: was replaced with isBitSet

angrygorilla (commented):
Inline notes on three issues from the full review: the overheating intersection (potential loss of a genuine overheating channel + divergence from `ProbeStatus`), an inconsistent control-device state the debounce can publish, and the debounce's event-driven latching. The rest of the review (public API breaks, wildcard imports, test-coverage gaps) is non-blocking.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/scanning/ProbeAdvertisingData.kt
```diff
@@ -123,7 +118,38 @@ internal class ProbeAdvertisingData(
...
+                    manufacturerData.sliceArray(OVERHEAT_RANGE)[0]
+                )
+                // Corroborate per-channel, not just "is anything overheating": the raw byte can
+                // have individual bits incorrectly set by the same firmware bug, so only report a
+                // channel as overheating here if the independent temperature-derived calculation
+                // also flags that specific channel. Otherwise a spurious bit on one channel could
+                // masquerade as the real overheating channel while the true one goes unreported.
+                OverheatingSensors(
+                    temperatureOverheatingSensors.values.filter { it in rawOverheatingSensors.values }
+                )
```
angrygorilla: **The intersection is asymmetric in the dangerous direction — it can drop a genuinely overheating channel.**

`ProbeAdvertisingDataTest` pins this explicitly: temperatures say T3 *and T4* are over threshold, the raw byte reports T1+T3, and the result is `[2]`. T4's real over-threshold reading is silently suppressed.

Two concrete risks:

1. **Firmware that doesn't populate byte 21.** Android never read this byte from advertising data before this PR, so there's no field evidence that every probe/repeater build writes it. If any build leaves it zeroed while temperatures are genuinely hot, advertising now reports *zero* overheating channels where it previously reported the full temperature-derived set. The `size > OVERHEAT_RANGE.last` guard doesn't help — the byte is present, just not populated.

2. **Divergence from the connected path.** `ProbeStatus.kt:127-138` does *not* intersect — when any temperature is over threshold it takes the raw byte wholesale. The same probe in the same state now reports different overheating sets depending on whether you're reading advertisements or status notifications.

The FW <= 2.2.0 spurious-bit bug this is defending against is already fully neutralized by the outer `isAnySensorOverheating()` guard plus using the temperature-derived set — the raw byte can't add anything trustworthy on top of that. Suggest just using `temperatureOverheatingSensors` in this branch:

```kotlin
val overheatingSensors = OverheatingSensors.fromTemperatures(probeTemperatures)
```

Or, if the raw byte is wanted for a reason I'm missing, at minimum align `ProbeStatus` and `ProbeAdvertisingData` on one policy so the two paths can't disagree.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/EngineManager.kt
```diff
@@ -150,38 +169,100 @@ internal class EngineManager(
...
+                    it.copy(
+                        engineBatteryStatus = status.engineBatteryStatus,
+                        engineStatusFlags = status.engineStatusFlags.copy(
+                            controlDeviceConnected = controlDeviceConnected,
+                        ),
+                        engineFanStatus = status.engineFanStatus,
+                        engineControllerStatus = status.engineControllerStatus,
+                        temperatureSetPointCelsius = status.temperatureSetPoint,
+                        controlDeviceType = controlDeviceType,
+                        controlSerialNumber = controlSerialNumber,
```
angrygorilla: **The debounce can publish `controlDeviceConnected = true` alongside a null controller identity.**

The `else` branch above debounces the *flag*, but these two lines write the incoming status's identity unconditionally. If firmware ever reports `controlDeviceConnected = false` **and** zeroed control-device bytes (-> `takeIfIsSet` -> `null`) while `previous.controlSerialNumber != null`, the engine publishes `controlDeviceConnected = true` with `controlSerialNumber = null` for up to 30 s.

That directly contradicts `Engine.hasConfirmedController` (`Engine.kt:79`), which requires both a non-null serial and type — so a consumer reading `engineStatusFlags.controlDeviceConnected` and one reading `hasConfirmedController` will disagree for the whole debounce window.

The app-initiated clear path is safe (`setControlDevice` clears `controlDeviceDisconnectPending` under the lock, so the next status takes the `previous.controlSerialNumber == null` branch). This is only reachable on a firmware-initiated clear — which is exactly the case `EngineStatus.fromRawData`'s new comment says it is *not* protecting against.

Cheapest fix is to keep the identity consistent with the debounced flag:

```kotlin
controlDeviceType = if (controlDeviceConnected) controlDeviceType ?: previous.controlDeviceType else controlDeviceType,
controlSerialNumber = if (controlDeviceConnected) controlSerialNumber ?: previous.controlSerialNumber else controlSerialNumber,
```

or alternatively fall through to the immediate-clear branch when `status.controlSerialNumber == null`, on the grounds that a cleared identity is itself confirmation there's nothing left to debounce.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/EngineManager.kt
```diff
@@ -39,10 +39,19 @@ import inc.combustion.framework.ble.scanning.EngineAdvertisingData
...
 import kotlinx.coroutines.flow.*
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+import kotlinx.coroutines.withContext
 import java.util.concurrent.atomic.AtomicBoolean
 
+// How long a control device must be continuously reported as disconnected before the engine is
+// treated as uncontrolled, so a brief radio drop doesn't flash the UI to "unattended."
+private const val CONTROL_DEVICE_DISCONNECT_CONFIRMATION_DELAY_MS = 30_000L
```
angrygorilla: **The confirmation is event-driven, so it can latch.**

`IdleMonitor.isIdle()` is only evaluated inside `handleStatus`, so the 30 s window only advances when a *new* status arrives. If the engine goes silent immediately after the first disconnected status — a fairly plausible shape for the radio drop this is meant to ride out — nothing ever re-evaluates and `controlDeviceConnected` stays `true` indefinitely. No timer fires it.

`statusNotificationsStale` covers the UI in practice, but the constant and its comment read like a wall-clock guarantee ("continuously reported as disconnected for 30 s"), which isn't what the code does. Worth either:

- noting the dependency here, e.g. *"only re-evaluated on status arrival; a fully silent engine is handled by `statusNotificationsStale` instead"*, or
- driving confirmation from a `scope.launch { delay(...) }` cancelled on reconnect, if the latch is actually reachable in the field.

Either way it'd be good to have a test for it — `EngineManagerTest` already mocks `SystemClock.elapsedRealtime`, so "disconnected status, then advance 60 s with no further statuses" is a couple of lines.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/scanning/ProbeAdvertisingData.kt
jhavatar: reverted unnecessary changes

combustion-android-ble/src/main/java/inc/combustion/framework/ble/EngineManager.kt
jhavatar: controlDeviceType and controlSerialNumber is not directly linked to controlSerialNumber despite name similarity. controlDeviceType and controlSerialNumber is the device that engine will attempt to connect to and the values do not change if engine is actually connected or not -- therefore not modified during debouncing

combustion-android-ble/src/main/java/inc/combustion/framework/ble/EngineManager.kt
jhavatar: Moved debouncing logic out of framework into app to also include cloud data in the processing. In app I included updating status when debounce times out with controlDeviceConnected == false.

angrygorilla (approved):
Nice work!

PR #185, opened 2026-06-24
DROID-757 : export CSV updates and other fixes

jhavatar:
## Description
- update csv export:
    - support gauge
    - add session start date to header
    - add a notes column
    - update version to 6
- Trim trailing null-byte padding from fixed-width UTF-8 fields: Fixed-width fields (serial numbers, hardware/firmware revision) are null-padded by firmware when the string is shorter than the field width -- utf8StringFromRange() didn't strip that padding. companion app PR: https://github.com/combustion-inc/combustion-android-prod/pull/418

### Note, app uses its own export and not framework's since framework is limited to ble data.

### DIFF
--- a/combustion-android-ble/src/main/java/inc/combustion/framework/Extensions.kt
+++ b/combustion-android-ble/src/main/java/inc/combustion/framework/Extensions.kt
@@ -49,8 +49,12 @@ fun UByte.setBit(bit: Int): UByte {
 
 fun UByte.toPercentage(): Int = this.toInt()
 
-fun UByteArray.utf8StringFromRange(indices: IntRange): String =
-    String(this.copyOf().sliceArray(indices).toByteArray(), Charsets.UTF_8).uppercase()
+// These are fixed-width fields; firmware pads them with trailing NUL bytes when the actual
+// string is shorter than the field width, so we trim those before returning.
+fun UByteArray.utf8StringFromRange(indices: IntRange, uppercase: Boolean = true): String =
+    String(this.copyOf().sliceArray(indices).toByteArray(), Charsets.UTF_8)
+        .trim('\u0000')
+        .let { if (uppercase) it.uppercase() else it }
 
 fun UByteArray.getUtf8SerialNumber(startIdx: Int): String =
     this.utf8StringFromRange(startIdx until (startIdx + UTF8_SERIAL_NUMBER_LENGTH))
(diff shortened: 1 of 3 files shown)


angrygorilla (approved):

combustion-android-ble/src/main/java/inc/combustion/framework/service/DeviceManager.kt
```diff
@@ -57,12 +57,17 @@ import kotlinx.coroutines.Dispatchers
...
+import java.time.Instant
 import java.time.LocalDateTime
+import java.time.ZoneId
 import java.time.format.DateTimeFormatter
 import java.util.Date
 import java.util.Locale
 import java.util.concurrent.atomic.AtomicBoolean
 
+private const val CSV_VERSION = 6
+private const val DEFAULT_SAMPLE_PERIOD: UInt = 5000u
```
angrygorilla, line 69: The fallback for a null `sessionInfo` changed from `?: 0` to `?: 5000u` here (and at the gauge call on line 1224). When session info is missing, the CSV header now reads `Sample Period: 5000`, which is indistinguishable from a genuine 5-second sample period — in fact `SimulatedProbeBleDevice` constructs `SessionInformation(0x12345678u, 5000u)`, so 5000 is a real value elsewhere in the codebase. The old `0` was an obvious out-of-band sentinel a consumer could detect; this replaces it with a plausible-but-wrong number precisely on the path where the metadata is least trustworthy.

Suggest reverting to `0`, or emitting `"UNKNOWN"` to match how the other null metadata fields (S/N, FW, HW) already render.

combustion-android-ble/src/main/java/inc/combustion/framework/Extensions.kt
```diff
@@ -49,8 +49,12 @@ fun UByte.setBit(bit: Int): UByte {
 
 fun UByte.toPercentage(): Int = this.toInt()
 
-fun UByteArray.utf8StringFromRange(indices: IntRange): String =
-    String(this.copyOf().sliceArray(indices).toByteArray(), Charsets.UTF_8).uppercase()
+// These are fixed-width fields; firmware pads them with trailing NUL bytes when the actual
+// string is shorter than the field width, so we trim those before returning.
+fun UByteArray.utf8StringFromRange(indices: IntRange, uppercase: Boolean = true): String =
```
angrygorilla, line 54: The new `uppercase` parameter has no production caller — only `ExtensionsTest` passes `uppercase = false`. The two sites that would use it, `NodeReadHardwareRevisionResponse.kt:68` and `NodeReadFirmwareRevisionResponse.kt:66`, still hand-roll `String(raw.toByteArray(), Charsets.UTF_8).trim('\u0000')`, which is now character-for-character what this extension does. Looks like an intended consolidation that stopped halfway: either point those two at `utf8StringFromRange(range, uppercase = false)`, or drop the parameter until there's a caller.

One thing to fold into that change: `.trim('\u0000')` strips leading NULs too, but the comment says trailing. A leading NUL would be an offset bug, and trimming it silently hides the symptom — `trimEnd('\u0000')` matches the stated contract and fails loudly instead. (The two hand-rolled sites have the same issue.)

PR #186, opened 2026-07-16
Feature/droid 749 framework supports engine

jhavatar:
## Desciption
- support Engine devices and sent ble messages to engines
- update kable ble lib to v0.42.0
- update compile SDK version to 36 -- required by kable and matches app SDK


combustion-android-ble/src/main/java/inc/combustion/framework/log/LogManager.kt
```diff
@@ -283,6 +283,10 @@ internal class LogManager {
         )
     }
 
+    fun manageEngine(scope: CoroutineScope, engineManager: EngineManager) {
+        //TODO: implement
```
angrygorilla: **Engine log transfer is a non-functional stub that hides a crash landmine.**

`manageEngine()` is an empty `//TODO: implement`, so unlike `manageGauge`/`manageProbe` it never registers the engine for log transfer: `uploadState` stays `Unavailable`, `recordsDownloaded`/`logUploadPercent` never move, and no records are ever transferred. All the log scaffolding on `Engine` and `LoggedEngineDataPoint.fromDeviceStatus` is dead as a result.

Meanwhile `EngineManager.sendLogRequest()` is `TODO("Not yet implemented")`, which *throws* `NotImplementedError`. Today the two stubs mask each other (nothing drives `sendLogRequest`), but the moment `manageEngine` is wired up this becomes a hard crash.

If log support is out of scope for this PR, that's fine — but please make it explicit, and make the `sendLogRequest` override degrade gracefully (log + return) rather than throw, so a future `manageEngine` implementation can't turn it into a crash.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/NodeHybridManager.kt
```diff
@@ -0,0 +1,410 @@
...
+        maxSequenceNumber: UInt,
+    ): S {
+        if (sessionInfo != info) {
+            logTransferCompleteCallback()
+            uploadState = ProbeUploadState.Unavailable
+            Log.i(LOG_TAG, "PM($serialNumber): finished log transfer.")
+        }
+        sessionInfo = info
+        val updated = _deviceFlow.value.withSessionInfo(info, minSequenceNumber, maxSequenceNumber)
+        _deviceFlow.update { updated }
```
angrygorilla: **`StateFlow.update { }` reads `_deviceFlow.value` instead of the lambda's `it`, which defeats the CAS retry.**

`update {}` may run its block more than once under contention and expects a pure function of `it`. Reading `_deviceFlow.value` inside (here `handleSessionInfo` computes `updated` from `.value` then does `update { updated }`, ignoring `it`) means a concurrent field change landing between read and write is silently dropped.

This pattern is spread across the shared base — `updateConnectionState`, `handleOutOfRange`, `fetchFirmwareVersion`, `handleAdvertisingPackets` — so it now affects both Gauge and Engine, plus two fresh instances in `EngineManager.handleStatus` and `setTemperatureSetPoint`. It's inconsistent even within this file: `fetchHardwareRevision`/`fetchModelInformation` correctly use `it`. Worth a sweep to make every `update {}` a pure function of its argument.

combustion-android-ble/src/main/java/inc/combustion/framework/service/DeviceManager.kt
```diff
@@ -101,6 +101,9 @@ class DeviceManager(
         const val MINIMUM_PREDICTION_SETPOINT_CELSIUS = 0.0
         const val MAXIMUM_PREDICTION_SETPOINT_CELSIUS = 100.0
 
+        const val MINIMUM_ENGINE_SETPOINT_CELSIUS = 65.0
+        const val MAXIMUM_ENGINE_SETPOINT_CELSIUS = 302.0
```
angrygorilla, line 105: **These engine set-point bounds are declared but never enforced.**

`MINIMUM_ENGINE_SETPOINT_CELSIUS` / `MAXIMUM_ENGINE_SETPOINT_CELSIUS` are never referenced — `setEngineTemperatureSetPoint` passes any `SensorTemperature` straight to the wire with no range check. The prediction-setpoint path *does* validate against its min/max (`setRemovalTemperatureForProbe`), so this is an inconsistency rather than just a missing nicety. Either validate/clamp against these constants in `setEngineTemperatureSetPoint`, or drop them.

combustion-android-ble/src/main/java/inc/combustion/framework/service/DeviceDiscoveryEvent.kt
```diff
@@ -64,6 +72,13 @@ sealed class DeviceDiscoveryEvent {
         val serialNumber: String
     ) : DeviceDiscoveryEvent()
 
+    /**
+     * Combustion engine with serial number [serialNumber] was removed.
+     */
+    data class EngineRemoved(
```
angrygorilla, line 78: **`EngineRemoved` is defined here but never emitted — engine removal actually fires `ProbeRemoved`.**

In `NetworkManager.unlinkDevice` (`NetworkManager.kt:1384`) the removal event is chosen with `is GaugeManager -> GaugeRemoved` / `else -> ProbeRemoved`. An `EngineManager` isn't a `GaugeManager`, so it falls into `else` and an unlinked engine emits `DeviceDiscoveryEvent.ProbeRemoved(serialNumber)`. Consequently this `EngineRemoved` event is dead code, engine-removal listeners never fire, and probe-removal listeners get a spurious event for a serial they don't own.

Fix in `unlinkDevice`: add `is EngineManager -> DeviceDiscoveryEvent.EngineRemoved(serialNumber)` above the `else`.

combustion-android-ble/src/main/java/inc/combustion/framework/log/LogManager.kt
jhavatar: Added more to comment, added a log that firmware doesn't support logs yet, and sendLogRequest logs + returns instead of throwing."

combustion-android-ble/src/main/java/inc/combustion/framework/ble/NodeHybridManager.kt
jhavatar: Good catch, confirmed and swept the whole file. Fixed handleSessionInfo (was the worst offender — fully ignored it, now uses updateAndGet { it.withSessionInfo(...) }), updateConnectionState, handleOutOfRange, and handleAdvertisingPackets to build off the CAS snapshot instead of re-reading _deviceFlow.value. Also fixed the two new Engine instances you flagged (EngineManager.handleStatus and setTemperatureSetPoint) the same way. Left fetchHardwareRevision/fetchModelInformation as-is since they were already correct.

combustion-android-ble/src/main/java/inc/combustion/framework/service/DeviceManager.kt
jhavatar: Fixed — setEngineTemperatureSetPoint now validates against MINIMUM_ENGINE_SETPOINT_CELSIUS/MAXIMUM_ENGINE_SETPOINT_CELSIUS and calls completionHandler(false) on out-of-range input, matching the existing setRemovalPrediction pattern.

combustion-android-ble/src/main/java/inc/combustion/framework/service/DeviceDiscoveryEvent.kt
jhavatar: Fixed — added is EngineManager -> DeviceDiscoveryEvent.EngineRemoved(serialNumber) to the when in NetworkManager.unlinkDevice, above the else.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/GaugeManager.kt
```diff
@@ -638,4 +200,67 @@ internal class GaugeManager(
...
+            )
+
+            _normalModeStatusFlow.emit(status)
+
+            if (!simulated) {
+                fetchDeviceInfo()
+            }
+
+            _deviceFlow.update {
+                _deviceFlow.value.copy(
```
angrygorilla: **Stale-read inconsistency with `EngineManager`.** This `update {}` block builds from `_deviceFlow.value.copy(...)` instead of the lambda's `it`; `updateDataFromAdvertisement` (line ~257) similarly mixes `current` with `_deviceFlow.value.baseDevice`. It's functionally safe (`StateFlow.update` re-runs the block on CAS failure and `.value` is re-read each pass), but this is exactly the pattern that was just cleaned up in `EngineManager.handleStatus`, where it reads cleanly from `it`. Worth aligning Gauge to the same form so it doesn't read as the bug you fixed.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/device/SimulatedNodeHybridBleDevice.kt
```diff
@@ -0,0 +1,187 @@
...
+
+    private var observeStatusCallback: (suspend (status: SpecializedDeviceStatus) -> Unit)? = null
+
+    private var observeRemoteRssiCallback: (suspend (rssi: Int) -> Unit)? = null
+
+    private var observeConnectionStateCallback: (suspend (newConnectionState: DeviceConnectionState) -> Unit)? =
+        null
+
+    init {
+        fixedRateTimer(name = "SimAdvertising", initialDelay = 1000, period = 1000) {
```
angrygorilla, line 100: **Timers are never cancelled.** Both `fixedRateTimer`s (`SimAdvertising` here and `SimStatus` at line 111) are started in `init` and never stopped — there's no `finish()`/cleanup hook, so a simulated device that's removed keeps firing forever. The single-timer version of this predates the refactor, but hoisting it here doubles it and centralizes it, so it's a good spot to hold the `Timer` references and `cancel()` them on `disconnect()`/removal. Simulated-only, so impact is limited to test/demo apps.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/device/SimulatedEngineBleDevice.kt
```diff
@@ -0,0 +1,152 @@
...
+        knobAngleTenthsDegrees = 0u,
+        chargingFault = EngineChargingFault.NONE,
+    )
+
+    override fun sendSetTemperatureSetPoint(
+        temperature: SensorTemperature,
+        reqId: UInt?,
+        callback: ((Boolean, Any?) -> Unit)?,
+    ) {
+        GlobalScope.launch {
```
angrygorilla: **`GlobalScope.launch`** here (and at line 140) mutates `temperatureSetPoint` / `controlSerialNumber`, which are read by the device's own `scope`-driven timer in `generateStatus`/`generateAdvertisement` — a benign data race, but `GlobalScope` is discouraged. Prefer launching on the injected `scope` so the work is tied to the device lifecycle.

combustion-android-ble/src/main/java/inc/combustion/framework/service/FirmwareVersion.kt
```diff
@@ -122,4 +122,25 @@ data class FirmwareVersion(
...
\ No newline at end of file
+}
+
+/**
+ * Returns whether this version is at least [minimum]. If [inclusive] is false, [minimum] itself
+ * is not considered sufficient and this version must be strictly greater. Returns null if this
+ * version is unknown (null), since support can't be determined; returns true if [minimum] is
+ * null, since there's nothing to require.
+ */
+fun FirmwareVersion?.isAtLeast(minimum: FirmwareVersion?, inclusive: Boolean = true): Boolean? {
```
angrygorilla, line 133: **Dead code.** `isAtLeast` (here) and `isOlderThan` (line 143) have no callers anywhere in the module (grep-confirmed), and `CombustionProductType.typeInt` (CombustionProductType.kt:78) is likewise defined but unused. If these are meant to firmware-gate the future engine-log feature, add a comment/TODO tying them to that; otherwise drop them to avoid untested public surface.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/device/NodeHybridBleDevice.kt
```diff
@@ -0,0 +1,192 @@
+/*
+ * Project: Combustion Inc. Android Framework
+ * File: HybridBleDevice.kt
```
angrygorilla: **Copy-pasted `File:` name in the license header** — says `HybridBleDevice.kt` but the file is `NodeHybridBleDevice.kt`. Same issue in a few other new files: `EngineFanState.kt` (`File: FanState.kt`), `EngineFanStatus.kt` (`File: FanStatus.kt`), and `EngineBatteryLevel.kt` (`File: EngineBatteryStatus.kt`). Trivial, but easy to fix in one pass while you're here.

angrygorilla (commented):
Automated review — 5 inline findings. All low/trivial severity; nothing blocking. Feature work is solid and well-tested; these are cleanups and consistency items.

combustion-android-ble/src/main/java/inc/combustion/framework/service/FirmwareVersion.kt
jhavatar: helper functions accessed from the app. useful to other people using the framework

combustion-android-ble/src/main/java/inc/combustion/framework/ble/GaugeManager.kt
jhavatar: Fixed — aligned Gauge with the pattern used in Engine/NodeHybridManager: handleStatus now does _deviceFlow.update { it.copy(...) } instead of reading _deviceFlow.value inside the lambda, and updateDataFromAdvertisement now builds off current.baseDevice instead of _deviceFlow.value.baseDevice.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/device/SimulatedNodeHybridBleDevice.kt
jhavatar: Fixed — added a finish() to SimulatedNodeHybridBleDevice that cancels both timers, made BleManager.finish() open, and had NodeHybridManager override it to also call simulatedDevice?.finish() on manager teardown. Didn't hook this into disconnect() — that's called for ordinary disconnects too (not just removal), and simulated devices are meant to keep advertising while disconnected, so cancelling there would've broken reconnect.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/device/SimulatedEngineBleDevice.kt
jhavatar: Fixed — both call sites now launch on the injected scope (widened SimulatedNodeHybridBleDevice's scope param from private to protected so the subclass can reach it) instead of GlobalScope. Checked SimulatedGaugeBleDevice for the same pattern — it doesn't have one.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/device/NodeHybridBleDevice.kt
jhavatar: updated comments in NodeHybridBleDevice.kt, EngineFanState.kt, EngineFanStatus.kt, EngineBatteryLevel.kt.

angrygorilla (approved):
Working great so far! Nice work!

PR #187, opened 2026-07-31
DROID-765 : Retry BLE Commands

jhavatar:
## Description
Adds automatic retry-with-confirmation support for BLE commands, replacing today's send-once-and-hope model. Introduces `CommandCoordinator`, a per-manager component that retries a command every 5s for up to 30s, completing early either from a matching response or from a device status update confirming the change already took effect — and wires it into `EngineManager`, `GaugeManager`, `ProbeManager`, and `WiFiNodesManager`.

## What changed

- **`CommandCoordinator`**: new coroutine-based retry/completion coordinator (`sendRoutedCommand`, `completeAttempt`, `confirmCommandStatus`, `handleDeviceDisconnected`). `valueConfirmation` builds an early-success predicate for "set field X to Y" commands that also treats "changed to something else" as done, so a competing command from another device doesn't cause indefinite retry-bouncing.
- **`UartBleDevice.MessageCompletionHandler`**: converted from a single pending-wait slot to a keyed dictionary, so a retry can register a new wait without clobbering an earlier attempt that's still outstanding.
- **Commands ported**: `EngineManager.setControlDevice`/`setTemperatureSetPoint`, `GaugeManager.setHighLowAlarmStatus`, `ProbeManager.setPowerMode`/`setProbeHighLowAlarmStatus`/`setProbeColor`/`setProbeID`/`setPrediction`/`configureFoodSafe`, `WiFiNodesManager.sendNodeRequestRequiringWiFi`. `sendNodeRequestRequiringWiFi` always passes `isConfirmed = { false }` — a generic consumer payload with no corresponding status field — and relies purely on a matching response.
- **Not ported** (documented, not silently dropped):
  - `ProbeManager.resetProbe` — Reset Thermometer (0x0A) starts a brand new cook session every time it runs; a spurious retry after a lost ACK (not a real failure) would start a second session on top of the first, silently orphaning whatever had already been logged. Not idempotent, so not safe to retry blindly — and there's no status field a reboot could confirm against anyway.
  - `ProbeManager.resetFoodSafe` — Reset Food Safe (0x08) *is* idempotent (resets counters to a fixed state), but still has no status field that fits `valueConfirmation`, and a spurious retry could still erase a few seconds of legitimately re-accumulated tracking. Left as a single-shot send rather than accepting that cost for a command with no way to confirm early anyway.
  - Log-transfer requests (streaming, not single-shot), `NetworkManager.silenceAllAlarms` (broadcasts to every connected device, doesn't fit a single-target coordinator), `WiFiNodesManager.sendFeatureFlagRequest` (internal bookkeeping, not consumer-facing).

## Bugs found and fixed along the way

- **`confirmCommandStatus` gap**: a command whose `send()` never registered an attempt key (nothing reachable) was invisible to status-based confirmation. Fixed by tracking all in-flight commands independently of their registered keys.
- **Probe direct-link response matching**: the raw probe-direct UART protocol has no request-ID concept — every response is matched against a fixed `null` key. Passing a real per-attempt ID (the naive "give every attempt a unique ID" approach) breaks completion entirely rather than just risking a retry collision. Fixed by always sending `reqId = null` on the direct-link path, and separately preventing the resulting retry-vs-internal-timeout race with a manager-specific retry interval kept strictly above the device's per-transmission timeout.
- **Concurrent duplicate commands**: two overlapping calls to the same command on the same probe could silently steal each other's completion, since `CommandAttemptKey.Direct` has no per-call discriminator. Fixed with a per-`(device, command type)` mutex serializing the whole retry cycle.
- **Simulated device races**: several `Simulated*BleDevice` implementations resolved their callback synchronously, which could race `CommandCoordinator`'s key registration and silently drop the completion. Fixed by dispatching them asynchronously, matching the existing convention elsewhere in the simulated device layer.

## Testing

Unit tests added/extended for all of the above, including a `ProbeBleDeviceTest` that drives the real UART response-parsing path (hand-built raw frames with correct CRC) rather than mocking it away, and a `WiFiNodesManagerTest` that exercises the real WiFi-node-discovery flow. Full suite passes.


angrygorilla (commented):
Reviewed the retry-with-confirmation work. The design is sound and the KDoc is genuinely excellent — the "not ported and why" notes on `resetProbe`/`resetFoodSafe`, the `PROBE_DIRECT_RETRY_INTERVAL_MS` derivation, and the null-reqId direct-link explanation all capture reasoning that would otherwise be lost. `ProbeBleDeviceTest` driving real UART frames instead of mocking the parser is the right call. Unit suite passes locally.

Inline comments below. The two I'd treat as blocking are both in `valueConfirmation`'s semantics — they cause commands to report SUCCESS in the exact scenarios this feature was built for:

1. A `null` starting value (field not yet populated) is treated as "changed" by the first status that carries a value.
2. Alarm-status confirmation keys on `tripped`/`alarming`, which the device changes on its own.

The rest are worth folding into the same pass but aren't blockers.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/CommandCoordinator.kt
```diff
@@ -0,0 +1,295 @@
...
+         * @param extractValue Reads the relevant value out of a device status, or null if this
+         * particular status doesn't carry it.
+         */
+        fun <T> valueConfirmation(
+            startingValue: T,
+            commandedValue: T,
+            extractValue: (SpecializedDeviceStatus) -> T?,
+        ): (SpecializedDeviceStatus) -> Boolean = confirmed@{ status ->
+            val observed = extractValue(status) ?: return@confirmed false
+            observed == commandedValue || observed != startingValue
```
angrygorilla: **Spurious SUCCESS when the starting value was simply unknown.**

The KDoc handles an absent *observed* value, but not an absent *starting* value. Every field the six ported `ProbeManager` commands confirm against is nullable with a `null` default:

```
Probe.kt:103-124  id, color, predictionMode, setPointTemperatureCelsius,
                  foodSafeData, thermometerPrefs, highLowAlarmStatus  — all `= null`
```

So when `_deviceFlow.value.X` is still null, the *first* status update carrying a non-null value satisfies `observed != startingValue` and completes the command as SUCCESS — whether or not it was ever delivered.

This isn't theoretical. `highLowAlarmStatus`, `foodSafeData` and `predictionMode`/`setPointTemperatureCelsius` are only ever populated from a full `ProbeStatus` (`ProbeManager.updateHighLowAlarms`), never from advertising data. A probe that is advertising but not yet connected has all three at null — which is precisely the "no route available yet, keep retrying until one appears" case this feature exists to handle. Command issued → link comes up → first `ProbeStatus` arrives → spurious SUCCESS, retry loop stops, command never landed.

`EngineManager.setControlDevice` has the same shape (`Engine.controlSerialNumber` defaults null).

Suggested fix: when `startingValue == null`, fall back to strict `observed == commandedValue`. The anti-bouncing rationale for the "changed at all" branch only holds when you actually knew the prior value.

```kotlin
val observed = extractValue(status) ?: return@confirmed false
if (startingValue == null) return@confirmed observed == commandedValue
observed == commandedValue || observed != startingValue
```

combustion-android-ble/src/main/java/inc/combustion/framework/ble/ProbeManager.kt
```diff
@@ -599,45 +863,84 @@ internal class ProbeManager(
...
 
-                } else {
-                    onCompletion(false)
-                }
+                        if (sent != null && key != null) setOf(key) else emptySet()
+                    },
+                    isConfirmed = CommandCoordinator.valueConfirmation(
+                        startingValue = startingHighLowAlarmStatus,
+                        commandedValue = probeHighLowAlarmStatus,
+                        extractValue = { (it as? ProbeStatus)?.probeHighLowAlarmStatus },
```
angrygorilla: **Confirmation keys on device-driven state, not just what the command set.**

`HighLowAlarmStatus.AlarmStatus` (HighLowAlarmStatus.kt:38-42) carries `tripped` and `alarming` alongside the commanded `set`/`temperature`, and this compares the whole data class by equality. `ProbeHighLowAlarmStatus` aggregates 11 of them, so an *unrelated* already-armed alarm on any other sensor tripping during the 30s window is enough.

When that happens, `observed != startingValue` fires and the command reports SUCCESS without having landed. Worth extracting only the fields the command actually sets (`set` + `temperature` per sensor) rather than the full struct.

Same issue in `GaugeManager.setHighLowAlarmStatus`.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/GaugeManager.kt
```diff
@@ -136,46 +140,73 @@ internal class GaugeManager(
...
-                }
-            } else {
-                onCompletion(false)
+
+                    if (sent != null) setOf(key) else emptySet()
+                },
+                isConfirmed = CommandCoordinator.valueConfirmation(
+                    startingValue = startingHighLowAlarmStatus,
+                    commandedValue = highLowAlarmStatus,
+                    extractValue = { (it as? GaugeStatus)?.highLowAlarmStatus },
```
angrygorilla: Same problem as flagged on `ProbeManager.setProbeHighLowAlarmStatus`: `HighLowAlarmStatus` carries device-driven `tripped`/`alarming` bits (HighLowAlarmStatus.kt:38-42) alongside the commanded `set`/`temperature`. If the alarm trips anywhere in the 30s window, `observed != startingValue` fires and the command is confirmed as SUCCESS even though it never landed.

Compare only the sub-fields this command sets.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/ProbeManager.kt
```diff
@@ -397,114 +432,288 @@ internal class ProbeManager(
...
+                                commandCoordinator.completeAttempt(key, success = true)
                             }
                         }
-                    }
 
-                } else {
-                    onCompletion(false)
-                }
+                        val sent = simulatedProbe?.sendSetProbeID(probeId, null, onResponse)
+                            ?: arbitrator.directLink?.sendSetProbeID(probeId, null, onResponse)
```
angrygorilla: **`arbitrator.directLink` is read twice per attempt** — once above to build the key, again here to send.

`DataLinkArbitrator.directLink` is a computed getter over `bleDevice?.isConnected` (DataLinkArbitrator.kt:48-52), so it can flip between the two reads. If it goes non-null → null, the send falls through to the node path with `nodeRequestId` while the registered key is `Direct(...)`, and that attempt's response can never match. It degrades rather than breaks (the next retry recovers), but using the already-captured `directLink` local for the send too is a one-liner:

```kotlin
val sent = simulatedProbe?.sendSetProbeID(probeId, null, onResponse)
    ?: directLink?.sendSetProbeID(probeId, null, onResponse)
    ?: run { ... }
```

Same pattern in `setProbeColor`, `setPrediction`, `configureFoodSafe`, `setPowerMode` and `setProbeHighLowAlarmStatus`.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/device/UartBleDevice.kt
```diff
@@ -51,45 +51,54 @@ internal open class UartBleDevice(
...
-            localCallback?.let {
-                it(result, data)
-                matched = true
-            }
-
-            return matched
+            // Remove before invoking the callback (not after) so a re-entrant wait() call from
+            // within the callback -- e.g. an immediate retry -- isn't rejected by a stale entry
+            // that's about to be cleaned up anyway.
+            val entry = pending.remove(Key(reqId)) ?: return false
```
angrygorilla: **Undocumented behavior change: a null-registered wait no longer matches a non-null response ID.**

The old `handled()` short-circuited on `requestId?.let { ... }` — if `requestId` was null it invoked the callback regardless of the incoming `reqId`. Keyed lookup means `Key(null)` and `Key(42u)` now never match.

No live caller trips this — every Node-routed `send*` path currently passes a non-null ID, and I checked them all. But the trap is loaded: these signatures are all `reqId: UInt?`, and `NodeRequest` auto-generates a random ID when handed null (NodeRequest.kt:202). So a future `sendSetPowerMode(mode, null, cb)` on a `RepeatedProbeBleDevice` would register `Key(null)`, transmit a random ID, and silently drop the response forever — no timeout distinction, no log.

The class KDoc already documents the *other* deliberate change (different reqIds coexisting); worth documenting this one alongside it, or adding a warn log when `handled()` finds no entry for a non-null `reqId`.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/device/UartBleDevice.kt
```diff
@@ -98,53 +107,33 @@ internal open class UartBleDevice(
...
+            // is already pending, reject the new one -- matching the original handler's rejection
+            // behavior for a repeated wait on the same key.
+            if (pending.putIfAbsent(key, entry) != null) {
                 callback?.let { it(false, null) }
                 return
             }
 
-            waiting.set(true)
-            completionCallback = callback
             requestId = reqId
```
angrygorilla, line 146: `requestId` is no longer cleared when a wait completes or times out — the old `cleanup()` reset it to null on both paths, and now only `cancel()` does.

`RepeatedProbeBleDevice.sendSessionInformationRequest` / `readProbeFirmwareVersion` / `readProbeHardwareRevision` / `readProbeModelInformation` all read `handler.requestId` back to build the outgoing request. They're guarded by `if (!handler.isWaiting)` so they're correct today, but a stale non-null `requestId` sitting on a non-waiting handler is easy to misuse.

Given the KDoc already concedes this field is "not a reliable *the* pending request ID", clearing it when `pending` empties would cost nothing and remove the trap.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/WiFiNodesManager.kt
```diff
@@ -67,24 +68,71 @@ internal class WiFiNodesManager(
...
-            val mutex = getNodeMutex(deviceId)
-            mutex.acquire()
-            it.sendNodeRequest(request) { status, data ->
-                mutex.release()
-                completionHandler(status, data as? GenericNodeResponse)
-            }
-        } ?: run {
-            completionHandler(false, null)
+        val mutex = getNodeMutex(deviceId)
+        mutex.acquire()
```
angrygorilla, line 122: **Fail-fast is gone here.** Previously `connectedWiFiNodes[deviceId] == null` returned `completionHandler(false, null)` immediately. Now the semaphore is acquired up front and the coordinator burns the full 30s before failing, even when nothing is connected and `send` can never register a key.

That's a real change for callers — any UI awaiting this now shows a spinner for 30s instead of failing instantly. Worth confirming with the app team, or short-circuiting when `connectedWiFiNodes[deviceId]` is absent *and* no node is expected to appear.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/WiFiNodesManager.kt
```diff
@@ -67,24 +68,71 @@ internal class WiFiNodesManager(
...
+                            responseData = data as? GenericNodeResponse
+                            if (success) {
+                                commandCoordinator.completeAttempt(key, success = true)
+                            }
+                        }
+
+                        setOf(key)
+                    }
+                },
+                isConfirmed = { false },
```
angrygorilla, line 162: **Opaque consumer payloads are now retried up to 6×.**

`request` is a caller-supplied `GenericNodeRequest` — the framework can't know whether it's idempotent, and `isConfirmed = { false }` means there's no confirmation signal to stop early either.

This sits oddly against the PR's own reasoning for *not* porting `resetProbe`/`resetFoodSafe`: a spurious retry after a lost ACK re-executes a command whose effects may not be idempotent. That argument applies at least as strongly to a payload the framework can't inspect at all.

`sendRoutedCommand` already has a `retriesEnabled` parameter that nothing currently uses — exposing it here as an opt-out (or opt-in) would resolve this cleanly.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/WiFiNodesManager.kt
```diff
@@ -67,24 +68,71 @@ internal class WiFiNodesManager(
...
+                        val freshRequest = GenericNodeRequest(
+                            outgoingPayload = request.outgoingPayload,
+                            nodeSerialNumber = request.nodeSerialNumber,
+                            requestId = requestId,
+                            payloadLength = request.payloadLength,
+                            messageId = request.messageId,
+                        )
+
+                        node.sendNodeRequest(freshRequest) { success, data ->
+                            responseData = data as? GenericNodeResponse
```
angrygorilla: `responseData` is last-writer-wins across attempts. An earlier attempt's per-transmission timeout callback fires `responseData = null` and can land after a later attempt already succeeded.

The success path itself is safe — the write happens before `completeAttempt` → `deferred.complete()`, which gives a happens-before edge to the awaiting coroutine — but the interleaving above isn't ordered by anything. Assigning only when `success` is true would close it:

```kotlin
node.sendNodeRequest(freshRequest) { success, data ->
    if (success) {
        responseData = data as? GenericNodeResponse
        commandCoordinator.completeAttempt(key, success = true)
    }
}
```

combustion-android-ble/src/main/java/inc/combustion/framework/ble/CommandCoordinator.kt
```diff
@@ -0,0 +1,295 @@
...
+     * changed the value to something else doesn't cause indefinite retry-bouncing between the two.
+     * @param retriesEnabled Whether [send] should be retried every [retryIntervalMs] until
+     * [requestTimeoutMs] elapses; when false, [send] is called once and the command waits out the
+     * full timeout for either completion source.
+     */
+    suspend fun sendRoutedCommand(
+        targetSerialNumber: String,
+        send: (attemptNumber: Int) -> Set<CommandAttemptKey>,
+        isConfirmed: (SpecializedDeviceStatus) -> Boolean,
+        retriesEnabled: Boolean = true,
```
angrygorilla, line 253: Nit: `retriesEnabled` has no production caller — only tests exercise `false`. Same for `completeAttempt`'s `success` parameter, which is never passed `false` outside tests.

Rather than dropping them, `retriesEnabled` looks like the right answer for `WiFiNodesManager.sendNodeRequestRequiringWiFi` (see the comment there on retrying opaque payloads).

.idea/misc.xml
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
+  <component name="ExternalStorageConfigurationManager" enabled="true" />
```
angrygorilla: Unrelated IDE churn — this also strips the `<?xml version="1.0" encoding="UTF-8"?>` declaration on line 1. Worth dropping from the PR.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/CommandCoordinator.kt
jhavatar: good catch but not all null values represent absent values though. to solve introduced
``` kotlin
internal sealed class ExtractedValue<out T> {
    data class Present<T>(val value: T) : ExtractedValue<T>()
    data object Absent : ExtractedValue<Nothing>()
}
```

combustion-android-ble/src/main/java/inc/combustion/framework/ble/ProbeManager.kt
jhavatar: Switch to comparing HighLowAlarmStatus.Threshold (set/temperature only) instead of the full status when confirming alarm-status commands in both ProbeManager.setProbeHighLowAlarmStatus and GaugeManager.setHighLowAlarmStatus

combustion-android-ble/src/main/java/inc/combustion/framework/ble/GaugeManager.kt
jhavatar: Switch to comparing HighLowAlarmStatus.Threshold (set/temperature only) instead of the full status when confirming alarm-status commands in both ProbeManager.setProbeHighLowAlarmStatus and GaugeManager.setHighLowAlarmStatus

combustion-android-ble/src/main/java/inc/combustion/framework/ble/ProbeManager.kt
jhavatar: fixed

combustion-android-ble/src/main/java/inc/combustion/framework/ble/device/UartBleDevice.kt
jhavatar: 👍 added both documentation and warning

combustion-android-ble/src/main/java/inc/combustion/framework/ble/device/UartBleDevice.kt
jhavatar: added back clearing MessageCompletionHandler.requestId when the pending map drains empty

combustion-android-ble/src/main/java/inc/combustion/framework/ble/WiFiNodesManager.kt
jhavatar: added back fail fast

combustion-android-ble/src/main/java/inc/combustion/framework/ble/WiFiNodesManager.kt
jhavatar: fair -- exposing retriesEnabled with default false.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/WiFiNodesManager.kt
jhavatar: makes sense -- updated

combustion-android-ble/src/main/java/inc/combustion/framework/ble/CommandCoordinator.kt
jhavatar: sendNodeRequestRequiringWiFi now exposes retriesEnabled in DeviceManager

.idea/misc.xml
jhavatar: removed misc.xml from repo

angrygorilla (approved):
Nice work!

PR #189, opened 2026-09-17
DROID-778: set/parse gauge id in framework

jhavatar:
## Summary

Implements `SET_GAUGE_ID` end-to-end (UART/MeatNet, `GaugeManager`, `DeviceManager`), letting the app assign a gauge's display ID the same way it already can for probes.

## What's new

- **UART/MeatNet**: `NodeSetGaugeIDRequest`/`NodeSetGaugeIDResponse` (opcode `0x63`, 11-byte payload: 10-byte serial + 1-byte id), wired through `NodeBleDevice`, `GaugeBleDevice`, `SimulatedGaugeBleDevice`, and `GaugeManager.setGaugeID`. Verified byte-for-byte against the `combustion-gauge` firmware source (`SetGaugeIDRequest`/`Response` in `ble_node_uart_message.hpp`).
- **`GaugeID`**: a new public enum, distinct from `ProbeID` — a gauge can be assigned an ID beyond the 8-entry picker (firmware places no upper bound on the byte), so this type stays honest about "not one of the 8" rather than silently reporting `ID1`. Shares its byte&lt;-&gt;enum lookup with `ProbeID` via a new `IdTag` interface.
- **Reading the id back**: `GaugeStatus.id` (parsed from the status byte at offset 24, right after hop count) and `GaugeAdvertisingData.gaugeId` (parsed from the advertising byte at offset 20, right after gauge preferences) — both verified against the firmware's `GaugeStatus`/`GaugeMfgData` structs. Both are backward-compatible: a byte-count check falls back to a default (`0`/`null`) for firmware that predates this field, instead of misparsing or crashing. `Gauge.knownId` resolves the raw byte to a `GaugeID?`, `null` meaning "not one of the 8," not "assume ID1."
- **`DeviceIdManager<T, D>`**: generalized from the probe-only `ProbeIdManager` so probe and gauge auto-conflict-resolution (two devices reporting the same ID) share one implementation. Type-specific behavior (which id to extract, how to rank a serial-number tie-break) is injected per instance; gauge conflicts only ever consider devices with a *known* id, per `knownId`'s nullability.

## Bugs fixed along the way

- `GaugeManager.handleStatus` wrote the parsed id to `_deviceFlow` *after* the update that flips `isPlaceholder()` to false, letting `DeviceIdManager` briefly observe a non-placeholder gauge with a stale id and act on it with a real hardware write.
- `GaugeStatus.RAW_SIZE` didn't cover the unconditionally-sliced hop-count byte, so an exactly-23-byte legacy packet crashed instead of being rejected.
- `GaugeStatus.isNewRecord` was read as a little-endian `UShort` from a 1-byte range, pulling in the following hop-count byte as the high byte — silently misreporting `false` for any multi-hop relayed gauge. Firmware confirms `newRecordFlag` is a genuine standalone byte.
- `SimulatedGaugeBleDevice.sendSetGaugeID` wasn't persisting the assigned id (unlike `SimulatedProbeBleDevice`'s equivalent), so a simulated gauge kept reporting `ID1` forever after a `setGaugeID` call.
- `IdTag`'s shared lookup used `enumValues<T>()`, which allocates a fresh array every call, regressing `ProbeID.fromRaw` from its original zero-allocation `when`; switched to the cached `.entries`.

## Testing

- `DeviceIdManagerTest` (generalized from `ProbeIdManagerTest`, plus new gauge-specific cases: unknown-id gauges never entering conflict tracking, releasing a stale assignment when a gauge's id becomes unknown, lexicographic tie-breaking).
- `GaugeStatusTest` (new): `isNewRecord` correctness across hop counts (including a regression case that fails against the pre-fix code), `RAW_SIZE` rejecting a too-short packet, id parsing with and without the trailing byte.
- `GaugeAdvertisingDataTest`: id parsing with and without the trailing byte.
- Full unit test suite green.


angrygorilla (commented):
A few issues, mainly around gauges on pre-GAU-96 firmware: their ID reads as 0 (ID1) rather than absent, which breaks the backward-compatibility fallback, ID conflict resolution, and `setGaugeID` confirmation. Details inline.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/scanning/GaugeAdvertisingData.kt
```diff
@@ -90,6 +100,12 @@ internal class GaugeAdvertisingData(
                 null
             }
 
+            val gaugeId = if (manufacturerData.size > ID_RANGE.last) {
```
angrygorilla: This size check never returns `null` for pre-ID firmware. Before GAU-96, `GaugeMfgData` had `uint8_t reserved[2]` right after `gaugePreferences`, and Shared commit `1ceb053` replaced it with `gaugeId; reserved[1]`, so the packet length didn't change. Old firmware also zero-inits the advertising data (`BleData::MfgData mfgData = {};` in `AO_Ble.cpp`), so byte 20 is always present and `0x00`.

Result: every gauge on older firmware parses `gaugeId = 0u` and shows up as ID1 instead of keeping its existing value. A length check can't distinguish old from new firmware here; that needs something else, e.g. the firmware version.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/GaugeStatus.kt
```diff
@@ -87,11 +98,23 @@ data class GaugeStatus(
...
+            // byte, so the comparison to 1 silently fails whenever hop count isn't HOP1 (whose top
+            // two bits are the only ones HopCount.fromUByte's mask/shift constrains to zero).
+            val isNewRecord: Boolean = data[NEW_RECORD_FLAG_RANGE.first].toInt() == 1
 
             val hopCount: HopCount = HopCount.fromUByte(data.sliceArray(HOP_COUNT_RANGE)[0])
 
+            // Older firmware doesn't send the trailing id byte at all -- default to 0 (displayed
+            // as ID1) rather than rejecting the packet, so status from those devices keeps parsing
+            // exactly as it did before this field existed.
+            val id: UByte = if (data.size > GAUGE_ID_RANGE.last) {
```
angrygorilla: Defaulting a missing ID byte to `0u` makes pre-ID firmware indistinguishable from a gauge that really is ID1. Unlike advertising, the status struct appends `gaugeId`, so it really is missing on old firmware, but this turns 'absent' into 'ID1'.

Downstream, `DeviceIdManager` has no filter for this, so an old gauge claims the ID1 slot. With one old gauge and one new-firmware gauge set to ID1, conflict resolution kicks in: either the new gauge gets moved off ID1 because of a gauge that can't hold an ID, or `SET_GAUGE_ID` is sent to the old gauge, which can't handle it. Suggest making `GaugeStatus.id` a `UByte?` (null when absent), with `handleStatus` keeping the existing id and `DeviceIdManager` ignoring gauges with no known id.

combustion-android-ble/src/main/java/inc/combustion/framework/ble/GaugeManager.kt
```diff
@@ -223,6 +223,78 @@ internal class GaugeManager(
...
+                            }
+                        }
+                    }
+
+                    if (sent != null) setOf(key) else emptySet()
+                },
+                isConfirmed = CommandCoordinator.valueConfirmation(
+                    startingValue = startingGaugeId,
+                    commandedValue = gaugeId.type,
+                    extractValue = { it.extractedAs<GaugeStatus, _> { s -> s.id } },
```
angrygorilla: Because `GaugeStatus.id` defaults to `0` on pre-ID firmware, this extractor always returns `Present(0)` for old gauges. `valueConfirmation` confirms on `observed == commandedValue`, so `setGaugeID(ID1)` on an old gauge (including one DeviceIdManager assigns during conflict resolution) reports success immediately, even though the firmware can't store an ID. This should return `ExtractedValue.Absent` when the status packet has no ID byte (falls out naturally if `GaugeStatus.id` becomes nullable).

combustion-android-ble/src/main/java/inc/combustion/framework/ble/DeviceIdManager.kt
```diff
@@ -0,0 +1,199 @@
...
+            return
+        }
+
+        val (lowerId, higherId) = if (allIds.indexOf(lowestAvailableId) < allIds.indexOf(conflictId)) {
+            Pair(lowestAvailableId, conflictId)
+        } else {
+            Pair(conflictId, lowestAvailableId)
+        }
+
+        // assign the losing (non-winning) serial number to the lowest id
```
angrygorilla: This comment is the reverse of what the code does: when `newDeviceWins` is true the new device gets `lowerId`, so the *winning* serial keeps the low slot. That matches the class KDoc and the original ProbeIdManager comment on develop ("assign highest serialNumber to lowest ProbeId"). Suggest: `// assign the winning serial number to the lowest id`.

combustion-android-ble/src/main/java/inc/combustion/framework/service/ProbeID.kt
```diff
@@ -50,17 +50,11 @@ enum class ProbeID(val type: UByte) {
...
-                0x05u -> ID6
-                0x06u -> ID7
-                0x07u -> ID8
-                else -> ID1
-            }
+            // fromUByte already masked this to 3 bits, so idTagFromType can never actually miss --
+            // ID1 here is just a defensive fallback, unlike GaugeID.fromUByte's genuine "not one
+            // of the 8" null (a gauge, unlike a probe, can be assigned an ID this enum doesn't
+            // model at all).
+            return idTagFromType(entries, raw.toUByte()) ?: ID1
```
angrygorilla: `fromRaw(raw: UInt)` is public, and `raw.toUByte()` narrows before the lookup, so values above 255 now wrap to a valid ID instead of hitting the ID1 fallback, e.g. `fromRaw(0x101u)` used to return ID1 and now returns ID2. Nothing in the framework passes values that large, but it's a silent behavior change for callers. Suggest guarding with `if (raw > 0xFFu) return ID1` or comparing against `raw` directly.

